PR #30, opened 2017-02-07
Add test for null value exclusion

vreuter:
Also simplify null value check in ctor.
Fixes https://github.com/epigen/looper/issues/29


vreuter:
Superseded by https://github.com/epigen/looper/pull/33

PR #33, opened 2017-02-10
Test and logging infrastructure

vreuter:
This establishes some package logging infrastructure and refactors testing system. The diff is a bit unwieldy in terms of files, but some of that may be do whitespace editor wonkiness. The main features to look at hear are changes to the tests and the creation of a basic logging system. There have also been some case-specific exceptions added, and the separation of some utility functionality. Those exceptions and utilities should have a few test cases, though, as the currently-in-place ones don't exercise them. There are some small file additions here that are used by tests and haven't been worked into a write-to-tempfile model yet as has been done with the test files for merge_table, sample annotations, and project config. I may want to shuffle some lines in the `conftest` but wanted to go ahead and put this up for some review if there's time/interest.


vreuter:
Oh and for tests, 40/40 pass, with one "expected failure," which is a test I went ahead and put in place to track the exception type for getitem-style access on AttributeDict as discussed in https://github.com/epigen/looper/issues/32

nsheff:
Nice. I'm looking through things. First question: why remove test project files (config and annotation files) and hard-code them in as strings? Then we are not testing the csv parsing...

vreuter:
I like to move the test data closer to the fixtures and assertions for clarity via naming & locality. Those lines are still written to & read from disk, just in a temporary directory created at test runtime and cleaned up at test completion. Right now those are set up as class-scoped fixtures, so the write-to-disk happens once for each requesting test class (2 in this case). Even though each test that needs a `Project` or `PipelineInterface` instance to do its work gets a fresh one since those fixtures are function-scoped, each such test reads from its enclosing class's corresponding file(s). I feel like we don't lose any test independence that way since the tests don't modify those config & data files. I could separate the data files write from the project config write, though, as is done with the pipeline interface file. Looking at it now it seems slightly strange to bundle those in the same fixture.

vreuter:
Thanks for reviewing by the way. Fire away with more questions & suggestions if/when they come up. Happy to go over it in-person, too.

tests/conftest.py
```diff
@@ -0,0 +1,314 @@
...
+def pytest_addoption(parser):
+    parser.addoption("--logging-level",
+                     default="WARN",
+                     choices=["DEBUG", "INFO", "WARN", "WARNING", "ERROR"],
+                     help="Project root logger level to use for tests")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def conf_logs(request):
+    setup_looper_logger(request.config.getoption("--logging-level"))
```
nsheff: Should we also make this an option for looper users, not just during testing?

looper/utils.py
```diff
@@ -0,0 +1,157 @@
...
+        # This will raise exception even if validation was attempted,
+        # but no sections were used. Effectively, delegate responsibility
+        # to the caller to initiate validation only if doing so is relevant.
+        if not self.section_to_fail_by_command:
+            raise ValueError("No commands validated")
+        return 0 == len(self.failures)
+
+
+
+def fails(command, name=""):
```
nsheff: Needs a better name; this is not tied to a class right now... a static function?

looper/utils.py
```diff
@@ -0,0 +1,157 @@
...
+import logging
+import os
+import yaml
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+
+class CommandChecker(object):
```
nsheff: will have to think about where this ultimately belongs... CommandChecker will perhaps belong with PipelineInterface... in `pipeline_interface` we add a list of required commands, which will be read/checked by CommandChecker. Just a thought

looper/models.py
```diff
@@ -5,1427 +5,1469 @@
 =======================
 
 Workflow explained:
-	- Project is created
-	- Add Sample sheet to project (spawns next)
-		- Samples are created and added to project (automatically)
+    - Project is created
```
nsheff: please keep space/tab consistent within project

looper/models.py
```diff
@@ -5,1427 +5,1469 @@
...
+            self.update_looperenv(looperenv_file)
+
+        # Here, looperenv has been loaded (either custom or default). Initialize default compute settings.
+        self._logger.info("Establishing project compute settings")
+        self.set_compute("default")
+        if self.compute is None:
+            raise ComputeEstablishmentException()
+
+
+        if self.DEBUG:
```
nsheff: Should this be updated to the logger method?

looper/models.py
```diff
@@ -5,1427 +5,1469 @@
...
+        self.add_entries(self.config)
+
+        # Overwrite any config entries with entries in the subproject
+        if "subprojects" in self.config and subproject:
+            self.add_entries(self.config['subprojects'][subproject])
+
+        # In looper 0.4 we eliminated the paths section for simplicity.
+        # For backwards compatibility, mirror the paths section into metadata
+        if "paths" in self.config:
+            print("Warning: paths section in project config is deprecated. Please move all paths attributes to metadata section.")
```
nsheff: logger?

looper/models.py
```diff
@@ -5,1427 +5,1469 @@
...
+
+        :param str looperenv_file: path to file with new looper
+            environment configuration data
+        """
+        try:
+            with open(looperenv_file, 'r') as handle:
+                self._logger.info("Loading %s: %s",
+                                  LOOPERENV_VARNAME, looperenv_file)
+                looperenv = _yaml.load(handle)
+                if self.DEBUG:
```
nsheff: again to logger?

looper/utils.py
vreuter: What's the motivation to nest it within another class? It seems like it could have use in isolation. The only reason that this is really a class is for several data structures to store results of the checks. I could see a case being made for a function version that just creates and instance and returns the `failed` result, since I think that'd be the most common use. The other data structures are just to provide more fine-grained inspection capability.

looper/utils.py
vreuter: I think this makes sense outside of a class since it's general-purpose. I guess it could go within `CommandChecker`, but if so then that should definitely remain a top-level class. Totally agree about the name, though; what do you think?

vreuter:
I switched other `print` statements to use the logging system and provided the logging config options as cmdl args in the `looper`. Thanks for pointing that out; I had intended to do that previously but forgot after adding it to the tests.

vreuter:
I started to readjust the whitespace as you mentioned, but I didn't I do much of that yet because I have some 💪 🤔 on tabs/spaces. I think 4 spaces is broadly preferred, and it's especially problematic for human and IDE eyes if tabs are 6 spaces not 4. PEP8 heavily favors spaces: https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces
While there's a caveat about consistency with existing, mixing the two and/or having tabs not be 4 makes style checkers difficult to use. Furthermore,  if we're working with external collaborators like snakemake at some point, I think it's likely there'd be a complaint about that. I'd vote (if I have one) heavily (if it's not binary) for switching to the standard 4 spaces.

nsheff:
Yes, yes, I know, I know... but "Tabs should be used solely to remain consistent with code that is already indented with tabs. " And this is such. Earlier PEPs did not specify spaces over tabs, btw. It was a disaster to allow them both from the beginning, but we are now stuck with it.

Personally, after extensive reading I have never understood the justification of the preference for spaces, and I've also found more people supporting (and more justification for) tabs, but I've always been willing to be convinced otherwise, and you are welcome to do so. The more important thing to me is keeping to a standard, and in this case we had one... we can change it, but we should do so deliberately.

nsheff:
Also PEP8 is **a** style guide, not **the** style guide :)

looper/utils.py
nsheff: I'm not suggesting nesting in another class, I was thinking of the module level. by *with* I meant *parallel to* . right now we're not even using this yet, right? it's still just a concept...

looper/utils.py
nsheff: well, maybe: it should be positive, not negative (*does command succeed?* not *does command fail?*) and then can be named, if standalone, as `is_command_callable`; if part of a `Command` class then it could be just `is_callable`

nsheff: in general I prefer positive statements (where True means it worked, False means not) rather than the other way around; I think people just think this way more naturally.

vreuter:
I don't have a strong preference I guess w.r.t. actual character(s) used I guess; I care most about readability. Since it sounds like you're open to persuasion (?), let me try! =D

1. Since most things have gone to 4 spaces (or 1 tab = 4 space), it's difficult to read things that deviate from that norm. It looks like 1 tab = 6 spaces in some/much of this.
2. Long lines are also bad; when they're broken up, it's nice to have the IDE visually align them. When tabs instead of spaces are used, that alignment often requires mixing them, which is ill-advised in 2 and **prohibited** in 3. Snakemake appears to use 3: https://snakemake.readthedocs.io/en/latest/getting_started/installation.html
3. I'd love it if everyone used an IDE and was meticulous with their style settings, but I've observed a general reluctance to use to what I think are great tools. As an alternative, style standardization can be accomplished by running code through checker(s) post-hoc. Flake8 and PyLint both use PEP8. It's easy enough to ignore whitespace warnings (I think some regarding function/class spacing disagree with what you may find in something like Code Complete), but doing so facilitates a boy-who-cried-wolf phenomenon.

looper/utils.py
vreuter: Ah OK, got it. Yeah, it was parallel to `PipelineInterface` before; it just felt like an easy candidate to split from `models` module since--as you said--it's unused. I could see it making sense in a `utils` from an external user's perspective (i.e., if someone wanted to use it for things other than `looper`), but yeah from our within-package perspective I agree--it'd make sense to put it back alongside `PipelineInterface`. I can do that as I bring individual classes out of `models`.

tests/conftest.py
vreuter: Forgot to note here, but yep, definitely! Done now.

looper/utils.py
vreuter: Yeah that's how it was previously, and I totally agree with that general principle. Reexamining, I agree here, too. I inverted the logic when I decided to track individual failures, but those could be stored as they are now while reverting the main API function of interest to a positive statement. Will make that change.

vreuter: Oooh, also there's a logic bug! Would've been squashed as soon as tests are written, but the way it is right now, the retval is inverted!

nsheff:
Tabs vs. spaces essentially comes down to *function* vs. *form*. Arguments for spaces (including yours) all revolve around form -- that is, the way the code looks. While I value pretty code, I favor tabs because I value function even more. In this case, another way to think of function is accessibility. An analogy is the web... new websites build pretty navigation systems, drop-down menus, etc; these look nice, but sometimes the prettiness (form) gets in the way of actually accomplishing something (function)... for instance, an older web browser that cannot render flash will leave those systems unable to browse some websites (then think of lynx, text-based retrieval, etc). The lowest denominator is also the most scriptable.

Spaces require special IDEs. While I use such IDEs, I value code that can be written and edited using the lowest common denominator... eg: vi, gedit, notepad. Spaces are incompatible with generic text editing. Until they build keyboards with a "4-spaces" button, I don't want to sit there typing 16 spaces for a 4-indented LOC -- on every line. Python is particularly problematic because of its reliance on whitespace. As such, a decision to use spaces is a decision to exclude editability by the most basic interface; in other words, a reduction in accessibility, and a requirement to move to more sophisticated software to access ability to write this code. Tabs preserve this accessibility. Therefore, tabs are superior for function. 

I acknowledge that spaces let you write prettier code. I love pretty code and I strive for that. I suppose I've just lived through enough garbage-looking code (which is particularly prevalent in biology) to tame down that desire a little bit. But I just philosophically think that while IDEs should be helpful and should be used,  writing code should not be extraordinarily difficult without them -- and I value this philosophy more than I value absolutely perfect-looking code. You can get close to pretty with tabs... but spaces will never solve the problem of granting access to the lowest-common-denominator text editor.

Let's talk more in person.

vreuter:
Thanks for reviewing @afrendeiro ! I can answer more about the first question than I can about the broader ones that you mentioned at the end. I think that the main reason for isolating the core data types defined in `models` is so that they can be packaged, installed, and used separately from `looper` as a whole (i.e., in/by `snakemake`). At the more fine-grained level, I think it'd be nice to split some of the core classes into their own modules and out of `models` for two reasons--one from our/dev side and one from the user side. Our reason is more ergonomic--personally, I don't have a *super* strong opinion I guess, but I think it's more pleasant to edit files that are a bit shorter. More important is the user side--imports of a subset of the ADTs now defined in `models` could be made to require an import of only the corresponding set of smaller modules than needing to access a single larger module. Essentially, if `models` splits from `looper`, that can become its own project/package that's used by `looper`, `snakemake`, and whatever else.

vreuter:
The inevitable, gnarly merge conflict in `models` is resolved now. I took care to preserve the logic from those fix branches @nsheff, especially with regard to derived columns, the `__dict__` copy, and the `pipelines_dir` . Tests still all pass but it'd be great to have an extra set of eyes on those sections just to make sure that I didn't miss anything. The relevant commits are  and 1370328361e58b2621fd8baa4237e4b70b6ab78a (which I initially missed in the merge). Thanks for referencing the issues in the commit messages; that made it easier for me to check!

vreuter:
The `models` and `looper` modules, as well as new ones that I've added should be using spaces. I didn't go in and switch all the other files yet--this PR is big enough--but will do so soon.

vreuter:
OK to merge now @nsheff ?

nsheff:
@afrendeiro Answers to a few of your comments:


> Thanks for this. A proper logging structure was long due, not to mention the tests.
> 
>     I've noticed you want to take some things out of models or to split models from looper. What's the motivation for this and how would that work?

Basically, `models` should be it's own module, which looper imports. `snakemake` and other tools will require models but not looper. so they are standalone, really. downstream analysis requires models but not looper. others could write other submission engines using models but not looper.
> 
>     Oh the historical fight of spaces vs tabs :) I will not bring any more arguments in favour of spaces but would just like to highlight how tabs are really not sustainable in the long run. Spaces are far more used and tabs are not supported in Python3. The two styles are just not compatible (e.g. try using pypiper.NGSTk functions in a script with spaces or with Python3). If we want users and other projects to use the looper API we'll have to switch to spaces. Now would be a good time.

Ok we will switch to spaces since that's what you two want.

> 
> More broadly, what are the next development plans? Is there a rough roadmap? Are we going to implement much more before launching a final 0.4?

0.4 is done; it's released. It is stable and useful already I will switch rc2 to final if there are no complaints. these updates are going into 0.5.

afrendeiro:
> 0.4 is done; it's released. It is stable and useful already I will switch rc2 to final if there are no complaints. these updates are going into 0.5.

Great I think that's the best option. Release 0.4 now and if needed, have a short cycle until the next version.
Still, I'm assuming you have a more broader plan for long term development or are most required features already here? It this case it would be good to collect ideas/requests/improvements in a roadmap for 0.5.

nsheff:
> Still, I'm assuming you have a more broader plan for long term development or are most required features already here? It this case it would be good to collect ideas/requests/improvements in a roadmap for 0.5.

I"m starting to reach out to others and get some feedback, so things will probably continue to develop. Roadmap is already in progress on the issues tab! feel free to submit ideas or comment there (please do!)

PR #34, opened 2017-02-14
Travis setup and spaces

vreuter:
Convert project setup script to spaces (I think that's the last .py file here with tabs) and establish TravisCI configuration. The automated build passes 40/40 tests on my personal fork with the 1 expected failure for the null value construction corner case of `AttributeDict`. I think it'd be awesome to to automate test builds for this project. With the TravisCI config file, I think that there's little--if anything--that we'd need to do other than grant Travis access to the epigen version of this looper. I granted access only for my personal fork of the project. This also sets lower bounds on dependency versions and separates test environment dependencies as an additional layer so that users don't need to have everything that the test environment does just to use looper.


vreuter:
Closing in favor of more comprehensive https://github.com/epigen/looper/pull/36

Issue #32, opened 2017-02-07
AttributeDict exception types


vreuter:
It seems a bit odd that `AttributeDict({})['a']` throws an `AttributeError` rather than `KeyError`. Is this to intimate to users that dot-notation is the preferred access method, to reinforce the functionality beyond `dict` that's being provided here, both, or just an implementation corollary? If it's the last of those, we could change wrap the `getattr` call in a try/except and raise `KeyError` in the exception case to make behavior align more with intuition.

nsheff:
Agreed; it's probably the last of these... but won't this resolve itself if AttributeDict extends dict instead of object? then getitem can just not be overridden.

vreuter:
Yep :bird: :bird: :gem:

vreuter:
Will be linked with https://github.com/epigen/looper/issues/31

Issue #31, opened 2017-02-07
AttributeDict dict-like methods


vreuter:
As I was trying to add test code, I noticed that things like making assertions about length for `AttributeDict` can't be done since it extends `object` and doesn't implement `__len__`. I'm new to the project, so maybe this is a deliberate design decision that was made? If not, though, I think that it could reduce confusion to implement things like `__len__`, and also comparisons. Currently, for instance, `AttributeDict({}) == {}` fails since `AttributeDict` doesn't implement `__eq__`. A way to simplify things like this may be to extend builtin `dict` and then override the behavior that we want to change. It seems like that inherit-and-override model would be more straightforward than recreating some of the builtin `dict` functionality, and it would please all of the POLA folks out there.

nsheff:
Not a deliberate decision; I think you're right that `AttributeDict` should extend `dict` instead of object.

PR #36, opened 2017-02-15
Features 0.5

### LINKED ISSUE #29: AttributeDict construction
The constructor, via `AttributeDict.add_entries`, ignores null values, causing tests for `AttributeDict` to fail when `None` is the value in the mapping that's the argument to the constructor.

vreuter:
First wave of 0.5 features
Closes #29 #31 #32 
Also adds numeric logging capability rather than level name, Travis configuration, finishes up some spacing stuff, and adds a few test inputs.
Metadata TODO items:
- [ ] Update `changelog`
- [x] Update branch for TravisCI


vreuter:
Travis build passes on my fork with the configuration provided here.

nsheff:
As an aside, maybe we should be merging changes like this into `dev` to leave the `master` branch as a release-only branch?

nsheff:
```
python tests/test_looper.py 
  File "tests/test_looper.py", line 13, in <module>
    from looper.models import COL_KEY_SUFFIX
ImportError: cannot import name COL_KEY_SUFFIX
```

nsheff:
We should probably keep the `changelog.rst` and `VERSION` files up-to-date at the PR stage

vreuter:
Definitely agree that they should be updated, and PR-stage rather than version-tagging stage probably makes more senses since the changes are still top-of-mind--why not save ourselves the trouble of trying to recall changes! I'll update those for these changes.

vreuter:
For the tests, you have to run with `pytest` rather than `python` to trigger the fixtures and such in `conftest` that are used by individual test modules.

vreuter:
The invocation is hierarchical, but the entire test suite here runs very quickly (68 tests in 1.00s when I just executed locally). But if you're trying to diagnose and remedy the cause of failing tests, it's nice that you can specify specific module/class/function.

nsheff:
ok great, I guess I just need to do a quick pytest tutorial.

vreuter:
This will go into `dev` instead

nsheff:
I don't think these need to be command-line adjustable: 
``[--logging-fmt LOGGING_FMT]
[--logging-datefmt LOGGING_DATEFMT]``

logging-level is fine, but these are overkill for a user and add some confusion I think... I see what you want to do but wonder if it needs to be at the CLI.

README.md
```diff
@@ -1,6 +1,7 @@
 # Looper
 
 [![Documentation Status](http://readthedocs.org/projects/looper/badge/?version=latest)](http://looper.readthedocs.io/en/latest/?badge=latest)
+[![Build Status](https://travis-ci.org/vreuter/looper.svg?branch=features_0.5)](https://travis-ci.org/vreuter/looper)
```
vreuter: Yeah, I pointed it to `master` now. Open to `dev` instead (or probably both is easy, too), if we want to display status for our active development instead/also. This is on my fork, though, not `epigen`, since I'm not sure if it's OK to allow Travis to access this code in `epigen`, and I don't think that I have sufficient permissions to grant it access if we do want to do that.

looper/looper.py
```diff
@@ -684,7 +646,29 @@ def main():
     if args.command == "run":
         if args.compute:
             prj.set_compute(args.compute)
-        run(prj, args, remaining_args)
+
+        # TODO split here, spawning separate run process for each pipelines directory in project metadata pipelines directory.
+        pipedirs = prj.metadata.pipelines_dir
+        _LOGGER.debug("Pipelines dirpath(s): {}".format(pipedirs))
+
+        for pipedir in prj.metadata.pipelines_dir:
```
nsheff: Instead of looping through pipeline directories in looper, the project object itself should know how to deal with them; someone outside of looper may need to handle multiple pipeline directories.

In general we should stick as much logic as we can into models, since it's universal.

looper/models.py
```diff
@@ -256,37 +375,90 @@ def __repr__(self):
         return "Project '%s'" % name + "\nConfig: " + str(self.config)
 
 
+    def finalize_pipelines_directory(self, pipe_path=""):
```
nsheff: Actually the vision should be that multiple directories can exist __simultaneously__ within the project and pipeline is chosen via this priority list at execution time; is this instead determining a single path at construction?

vreuter: No, it's setting a list as the attribute

nsheff:
line 877

```
                _LOGGER.debug("Added {} pipeline dirpath(s) to sys.path: {}".
                              format(len(pipeline_dirpaths), pipeline_dirpaths))
```

is yielding:
```
models:878 [DEBUG] > Added 25 pipeline dirpath(s) to sys.path: /home/ns5bc/code//ATACseq 
models:878 [DEBUG] > Added 25 pipeline dirpath(s) to sys.path: /home/ns5bc/code//ATACseq 
models:878 [DEBUG] > Added 25 pipeline dirpath(s) to sys.path: /home/ns5bc/code//ATACseq 
models:878 [DEBUG] > Added 25 pipeline dirpath(s) to sys.path: /home/ns5bc/code//ATACseq 
models:878 [DEBUG] > Added 25 pipeline dirpath(s) to sys.path: /home/ns5bc/code//ATACseq 
```

Issue #37, opened 2017-02-15
looper version


nsheff:
`looper -h` should show what version it is. what's the best way to do that?

nsheff:
Solved in current master branch for 0.5

PR #39, opened 2017-02-15
Version text

vreuter:
Display version text for --version and -V
Addressing #37


vreuter:
Left version as 0.4; that's what this is #37 is targeted for, yeah?

vreuter:
Just merged a subbranch on my fork that appears to fix https://github.com/epigen/looper/issues/37

nsheff:
probably 0.5... I consider 0.4 done

setup.py
```diff
@@ -29,8 +29,8 @@ def get_static(name, condition=None):
 # scripts to be added to the $PATH
 scripts = get_static("scripts", condition="'.' in x")
 
-with open("VERSION", 'r') as versionfile:
-	version = versionfile.read().strip()
+with open("looper/_version.py", 'r') as versionfile:
```
nsheff: is this really the best practice? Open a python file and parse it to extract the version? There must be a better way...

nsheff: can't we just use `looper.__version__` ?

nsheff: e.g. https://www.python.org/dev/peps/pep-0396/#classic-distutils

vreuter: I left it this way because the way in place was file-parsing-based. From what I could glean, there's a lack of consensus. I have absolutely no preference about this. If placing we want to go the placing it in `__init__.py` and then import as you mention, I'm totally on board with that. Mostly, I want to have these changes in master and the 0.4 version finalized so that I can stop worrying about mege conflicts arising.

vreuter:
Thanks @nsheff  for plotting with me for a few minutes on how to do this. Better to see this now:
```
└─ $ ▶ python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
>>> import looper
>>> looper.__version__
'0.4.0'
>>> quit()
```

nsheff:
perfect; but I was expecting you would leave the `__version__` definition in its own file, rather than moving to `init.py` -- fine with me this way but it's nice to have the version definition in its own named file so it's easier to find, and I don't think it ruins any of this

nsheff:
actually now that I look back at it it looks like you already had it right and I was confused...

vreuter:
Agreed! It's in its own file now while we still get the benefit of the `__version__` attribute on the imported module object.

vreuter:
I think that this should be all set now, can you confirm @nsheff ? I think we hashed out a good way to do the versioning; I like this way of handling it. We get:
1. Preservation of information in its own file for more intuitive access.
2. Import of __version__ info in main looper driver module for version and help text formatting.
3. Evasion of potential issues with package installation that PEP 0396 mention can arise when `setup.py` performs package imports.
 I merged the changes from epigen dev into this branch on my personal fork, so the fixes that went into master are included here.

nsheff:
Yes, this now works on both command line and interactive.

Issue #45, opened 2017-02-16
Column derivation is stuck on data_source


vreuter:
Reported by @nsheff 
https://github.com/epigen/looper/issues/40#issuecomment-280398632

vreuter:
Suspected resolved as of https://github.com/epigen/looper/pull/46

Issue #41, opened 2017-02-16
0.4-rc2 fails for looper_args pipelines


nsheff:
I think a fix on the looper_args: False thing broke the typical pipelines in rc2. Looper args are not being passed through even when the pipeline interface requests them.

PR #46, opened 2017-02-16
No more null logger

vreuter:
The test case that you provided now runs through on Rivanna @nsheff as a "Dry run" for ATAC and skipping other protocols since they're not defined in the associated protocol mappings file. I had to copy a few project files, tweak some paths and variables in the config, and copy over the data files to run this through from my account, as there were some write/read permissions issues since some items were referencing your user paths. Thus, there are several failure points so I'd be more comfortable with this if you can run that test case as you were earlier and validate it against your expectation after a `pip install --upgrade` from my fork's `fix_master` branch.
Fix https://github.com/epigen/looper/issues/44


vreuter:
Other errors that cropped up (once past the null logger issue) that should now be fixed are floating-point population of a format string and an attempt to pickle logger objects attached to `AttributeDict` subclasses.

nsheff:
were you going to incorporate the fix to #41 since this is aimed at 0.4?

vreuter:
Yep forgot about that, done now

nsheff:
Now it does complete, so some stuff is fixed. 

But there are a few things still...

- The logger messages appear to not be recognized by `less` the way I expected...does it have to do with outputting to stderr vs stdout?
- it's not spitting out the right information that it used to; e.g. 
`2017-02-16 21:51:25,906 looper.looper looper : 236 - [ERROR] > 
Not submitted: Required input files not found` 
It used to say which files those were.

And also, the input files should be found, so this hasn't solved #45, for example.

nsheff:
Ok scratch that...I was using the wrong config file... it appears to be working correctly....

nsheff:
Ok, yes I can confirm that it's working correctly now for at least that test case.

But you probably noticed how much harder it is to see what's going on with all the verbose logging :)

nsheff:
So I actually believe this is solving #45 now. Not sure what happened (maybe it was my error?)

vreuter:
Oh, excellent! I was going to ask for a bit more on the columns issue because I wasn't entirely sure what you meant. Maybe there's another case or two we could try tomorrow to confirm, or are you fairly confident that issue's all set now?

vreuter:
Going ahead and merging so that `master` is 🙏 repaired. Yes, agreed on verbosity ;D

PR #76, opened 2017-03-08
setup() args

vreuter:
While trying out the Makefile strategy from @nsheff for simplifying the process of getting files onto a remote host, I discovered that we were leaving open the possibility for a setup() error.
`install_requires` could've been duplicated as a keyword argument, but only if using Python < 2.7. I noticed this because the base version on the host on which I was installing is 2.6, and previously I had always installed after making 2.7 the priority version. 

These changes should match the keyword through which the dependencies are passed to `setup()` as specified differently by `setuptools` vs `distutils` and avoid duplication. This also removes the check for version <= 2.6; I think we have plenty of dict comprehensions scattered around the code, so 2.6 would fail on those.

Original error:
```
└─ $ ▶ make loopcode
rsync -av --exclude ".git/*" ~/code/looper/* $REMOTE:~/stage/looper/
Password: 
/home/vpr9v/.bashrc: line 7: module: command not found
/home/vpr9v/.bashrc: line 8: module: command not found
/home/vpr9v/.bashrc: line 9: module: command not found
sending incremental file list
setup.py

sent 5,166 bytes  received 227 bytes  1,198.44 bytes/sec
total size is 2,378,004  speedup is 440.94
ssh $REMOTE -t 'pip install --user --upgrade --no-deps ~/stage/looper'
Password: 
/home/vpr9v/.bashrc: line 7: module: command not found
/home/vpr9v/.bashrc: line 8: module: command not found
/home/vpr9v/.bashrc: line 9: module: command not found
Processing /home/vpr9v/stage/looper
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/tmp/pip-5f86Rr-build/setup.py", line 64, in <module>
        **extra
    TypeError: setup() got multiple values for keyword argument 'install_requires'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-5f86Rr-build
You are using pip version 7.1.2, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Connection to interactive.hpc.virginia.edu closed.
Makefile:6: recipe for target 'loopcode' failed
make: *** [loopcode] Error 1
```


requirements.txt
```diff
@@ -1,3 +1,3 @@
-colorama>=0.3.7
-pandas>=0.19.2
-pyyaml>=3.12
+colorama==0.3.7
+pandas
+pyyaml==3.11
```
nsheff: What does specifying an exact here instead of using `>=` do, exactly?

nsheff:
So -- looper doesn't work on 2.6; what exactly is this changing? I lost the big picture...

vreuter:
Duplication of kwargs to setup function call, and use of abstract not concrete reqs for project dependencies in test builds

vreuter:
Snapshots a working version

requirements.txt
vreuter: Sorry the working version snapshot comment should've been here

vreuter: I'm also fine with just ditching the requirements files entirely. The only ones really being used now are the dev & test ones, just for the Travis build. Those could just as easily be enumerated directly in the Travis config.

vreuter:
Also, I think that without this, installation will fail in a clean virtualenv. Or anywhere that colorama isn't installed.

afrendeiro:
Yes, v0.5-rc1 fails in a clean virtualenv because it's missing colorama. These changes in setup.py should fix that.

Although pip doesn't use the requirements.txt files, it would be nice if we could keep them. It's quite standard practice and helps to quickly see what are the requirements a certain program has straight from github for example.

nsheff:
@afrendeiro fine with me.

Alright, let's bump this to v0.5-rc2 and and merge into master whenever you're ready @vreuter

vreuter:
Yeah I thought about the reqs files a bit more last night & am totally with you @afrendeiro ; definitely nice to be able to have an easy access point for seeing what will be installed and adding anything if needed. I made setup.py just parse from there so we don't have to keep separate lists.

nsheff:
you forgot to update version/changelog :)

PR #82, opened 2017-03-14
Ensure that the models logger has a handler.

vreuter:
Came up when generating some data regarding https://github.com/epigen/looper/issues/79
There are two problems with the way that the current `models` logger behaves when `models` or one of its classes is imported (say, in iPython) rather than by an invocation of `looper` itself:

- The 'no handlers' message from `logging` (annoying, not so bad)
- No log messages are printed from the module in the iPython session (sort of bad)

This solves those problems.


nsheff:
If this isn't really critical, I think we should rather just call 0.5 done and merge this into dev to start 0.6 ...

vreuter:
Yep sounds good

PR #91, opened 2017-04-07
Alias test execution on package setup

vreuter:
Rather than requiring users to run `python setup.py pytest`, allow the more familiar `python setup.py test` command to trigger `pytest` just the same.


vreuter:
Motivated initially by https://github.com/epigen/looper/issues/89

PR #94, opened 2017-05-02
Implement implied columns

nsheff:
This PR implements a new concept called `implied_columns`. Implied columns are a solution to the following problem: I have a pipeline that needs to use 3 different arguments (strings) that are all based on a single input (in this case, it's organism). If my sample is from human, I need to set `genome`, `macs_genome_size`, and `prealignments`-- but all of these should be adjustable at the project level, so I can't hard-code into the pipeline just a single argument that then can construct these parameters all from that argument.

It makes most sense to just leave only 1 column in the annotation sheet, but then have the project specify additional columns that are implied by that 1 column. Thus, it's all managed by the project config file, and saving clutter in the annotation sheet.

This is entirely optional and doesn't change any existing functionality, just adds a new option to simplify more complex pipelines like this.

To do:
* [x] add more robust test cases
* [x] use more informational logger messages
* [x] add to documentation
* [x] add to changelog

### DIFF
--- a/tests/interactive.py
+++ b/tests/interactive.py
@@ -6,7 +6,8 @@
 
 from __future__ import absolute_import
 import looper
-from . import conftest
+#from . import conftest
+import conftest
 
 print("Running interactive.py tests")
 proj, pi = conftest.interactive()
--- a/tests/interactive.py
+++ b/tests/interactive.py
@@ -4,10 +4,7 @@
 # cd $CODEBASE/looper/tests
 # ipython
 
-from __future__ import absolute_import
-import looper
-#from . import conftest
 import conftest
 
-print("Running interactive.py tests")
+print("Establishing Project and PipelineInterface for testing and exploration")
 proj, pi = conftest.interactive()
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -1,6 +1,14 @@
 Changelog
 ******************************
 
+- **v0.6** (*unreleased*):
+
+  - New
+
+    - Adds support for implied_column section of the project config file
+
+    - Adds support for Python 3
+
 - **v0.5** (*2017-03-01*):
 
   - New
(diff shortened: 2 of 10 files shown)


tests/interactive.py
```diff
@@ -6,7 +6,8 @@
 
 from __future__ import absolute_import
 import looper
-from . import conftest
+#from . import conftest
+import conftest
```
nsheff: @vreuter This change is unrelated; but it was required -- is it because of changes way we're doing the 3 to 2 thing now?

vreuter:
1 & 2 on the checklist are addresses in https://github.com/epigen/looper/pull/95

tests/interactive.py
vreuter: Yeah I think that was the genesis but with the directory location mentioned in the comments it seems to be working with the bare/non-relative `conftest` import on both 2 and 3

nsheff:
testing complains of missing `mock`, but it is installed:
```
nsheff@puma:~/code/looper$ pytest
======================================================= test session starts =======================================================
platform linux -- Python 3.5.2, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: /home/nsheff/code/looper, inifile: pytest.ini
collected 29 items / 1 errors 

============================================================= ERRORS ==============================================================
______________________________________________ ERROR collecting tests/test_models.py ______________________________________________
ImportError while importing test module '/home/nsheff/code/looper/tests/test_models.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_models.py:6: in <module>
    import mock
E   ImportError: No module named 'mock'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================================== 1 error in 0.09 seconds =====================================================
nsheff@puma:~/code/looper$ python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
>>> import mock
>>> 
```

PR #95, opened 2017-05-03
Implied columns tests and messaging

vreuter:
TL;DR --> adds implied columns parsing context and tests and slightly improves existing functionality.

This adds some context to the messaging around implied columns parsing. It also adds tests for this new feature. Along for the ride is the removal of the restriction to `Series` on type of input data to `Sample` constructor, and ensuring that exception type reflects the access mode for a `Sample` data item. That is, dot-notation produces `AttributeError` and getitem syntax produces `KeyError` when the name of the data item requested isn't defined for the `Sample` instance in question. There's a small handful of tests here for those changes as well. Other changes are stylistic.

### DIFF
--- a/looper/__init__.py
+++ b/looper/__init__.py
@@ -41,7 +41,6 @@ def setup_looper_logger(level, additional_locations=None, devmode=False):
     :return logging.Logger: project-root logger
     """
 
-    logging.addLevelName(0, "EVERYTHING")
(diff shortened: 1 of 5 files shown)


vreuter:
Ticking off checklist tasks 1 & 2 from https://github.com/epigen/looper/pull/94

Issue #93, opened 2017-04-24
Make project definition more generic


johanneskoester:
Hi guys,
I have played with the API and must say I like it. Everything is clean and intuitive.
I think it should be generalized a bit though.

* Why does the sample sheet require a library column? This term may not apply in my experiment (or I don't care because it is all the same) and it seems a bit artificial to require that.
* Why the requirement of a data source/file column?
* The metadata section should not need the two keys output_dir and pipelines_dir, because they are not needed as well in every setting (e.g. when not using looper).
* I have a problem with the column `run` because it is ambiguous with what people sometimes use to call a sequencing run. Instead one could call it `blacklist`.

I really would like to see the sample sheet implementation together with these data source patterns outside of looper (without the looper specific syntax).

* [x] move `library` column requirement to looper. See #59 
* [x] move data source column requirement to looper
* [x] move `output_dir` and `pipelines_dir` requirements to looper
* [x] rename `run` to `toggle`

vreuter:
Here we have some specific questions and ideas related to https://github.com/epigen/looper/issues/74

vreuter:
Thanks @johanneskoester for trying this out and providing some feedback and ideas.

nsheff:
Agreed. Essentially we should move all these checks into looper and out of models.

* [ ] move `library` column requirement to looper. See #59 
* [ ] move data source column requirement to looper
* [ ] move `output_dir` and `pipelines_dir` requirements to looper

For the `run` column... I don't like `blacklist` because it, too, could be a common parameter name to a pipeline, messing things up. What about `toggle` or `runbool` instead?

Let's do all these things for milestone `0.6`

For separation, we just refer back to #74; continue discussion in that thread.

johanneskoester:
I like toggle.

afrendeiro:
Why not exactly what it means: `submit_job`?

johanneskoester:
Because this is not what it necessarily means in any context. For example, there could be only one job aggregating over all samples. In Snakemake, such sample metadata can be used in an arbitrary way.

afrendeiro:
Yes, I guess I should have said: "why not exactly what it means *for looper*".
In that case I guess `toggle` would accommodate both.

vreuter:
[...]

PR #98, opened 2017-05-08
More general models

vreuter:
Addresses/tests the first few 3 items in https://github.com/epigen/looper/issues/93.
I think that any checks for `library` are confined to the looper portion of the code. `data_source` is no longer appended to the `models` `Project` as a derived column, and the attempt to achieve other separation is implemented by subclassing the `models` `Project` with a few overrides for the `looper` specific version. Some exceptions have been replaced with builtins so that `models` isn't importing from `looper.exceptions`.


vreuter:
@nsheff perhaps you can run a few of your projects as test cases, as I've probably broken something that's not covered by the automated tests. @afrendeiro I tinkered a bit with the way in which `Sample` subclasses are handled, but not a ton so hopefully that's still in working order. If you have time to check that out let me know if you have some thoughts on it.

vreuter:
Now closes #93 I think, with the fourth item (`run` --> `toggle`) now included.

nsheff:
Can you describe in more detail "subclassing the models Project with a few overrides for the looper specific version." -- what exactly do we need to do in `looper` that doesn't fit in the generic `Project` that necessitates a subclass?

vreuter:
Yeah sure, maybe I should write a little of this in the docs and/or docstrings, too, for clarity and so we remember down the road. I think mostly overrides it's motivated by the separation broadly, and somewhat more specifically by the tasks in https://github.com/epigen/looper/issues/93. For instance `required_metadata` and `project_folders` are redefined for the `looper`-specific presumption of `output_dir`, the constructor specifies default environment settings that we have in `looper`, and the name inference and sample sheet addition methods tweak the base `models` versions to match the way we typically structure `looper` projects and use `derived_columns` / "data_source," respectively, that I think should not be relied upon in the more general `models` version.

nsheff:
Ok, makes sense -- I think `derived_columns` should be pretty universal, though, no?

vreuter:
Agree. Let me see if I can work this back into general modes.

vreuter:
OK, derived_columns has moved back to general `models`.

nsheff:
Ok thanks. A few other comments:
 * can you make a round of docs updates? in particular the run -> toggle thing should be in docs.
* why did you remove `data_source` ? 
> data_source is no longer appended to the models Project as a derived column

vreuter:
Yeah definitely, will do. Is it still removed as of most recent commit? I had taken it out but it should've returned now.

nsheff:
I didn't notice, sorry -- not sure

vreuter:
No worries, I've just added a couple cases to test that. The docs are updated a bit now, too.

PR #106, opened 2017-05-26
Skips by reason (not sample), testing recent changes

vreuter:
I had changed `looper.run`'s reporting of skip reasons to be sample-based rather than reason-based. Discussion with @nsheff revealed the fact that this is redundant since we already get that information on-the-fly from logging. This reverts to the previous reporting strategy (by skip/fail reason, not sample) and tests some code sections with recent changes, particularly a project config's `pipeline_args` and handling of defaults.


vreuter:
Motivated by https://github.com/epigen/looper/issues/102

nsheff:
looks great!

Issue #27, opened 2017-02-02
tsv sample annotations files


nsheff:
looper will choke on a tsv and give an incoherent error: (sample_name column doesn't exist).
instead it could probably be made to parse a tsv as well as a csv, or at least warn that the tsv needs to switch to a csv.

afrendeiro:
We could simply switch to pandas.read_table to read the annotation sheet which tries to determine the separator of a file. The read_csv and read_table functions are almost identical.

Optionally/additionally, we could implement a new CLI option that specifies the separator character and would be passed on: `pandas.read_csv(..., sep=args.sep)`.

vreuter:
I think the main difference between the pandas parsers is with regard to default header parsing behavior. We may have to take care to make sure we get the desired header handling, but yeah it'd be nice to just get the separator inference that way.

nsheff:
I would rather it just automatically guess either tsv or csv without having to specify.
I think if the `read_table` idea gets both csv and tsv headers right, we should do that.

PR #113, opened 2017-06-01
allow use of tsv format for samples

ninjha01:
Should solve #27

### DIFF
--- a/looper/models.py
+++ b/looper/models.py
@@ -831,7 +831,11 @@ class Project(AttributeDict):
             if self.metadata.merge_table is not None:
                 if _os.path.isfile(self.metadata.merge_table):
                     # read in merge table
-                    merge_table = _pd.read_csv(self.metadata.merge_table)
+                    name, extension = _os.path.splitext(self.metadata.merge_table)
+                    if extension == ".tsv":
+                        merge_table = _pd.read_csv(self.metadata.merge_table, sep='\t')
+                    else:
+                        merge_table = _pd.read_csv(self.metadata.merge_table)
 
                     if SAMPLE_NAME_COLNAME not in merge_table.columns:
                         raise KeyError(
@@ -948,8 +952,8 @@ class SampleSheet(object):
     """
     Class to model a sample annotation sheet.
 
-    :param csv: Path to csv file.
-    :type csv: str
+    :param path: Path to csv file.
+    :type path: str
     :param dtype: Data type to read csv file as. Default is str.
     :type dtype: type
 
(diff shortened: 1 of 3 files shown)


nsheff:
This relies on file extension to be accurate. Isn't the solution proposed in #27 superior as it will automatically get it right, regardless of file extension?

ninjha01:
Read_table caused errors in testing that I couldn't easily debug, and upon looking at 
[the documentation...](http://pandas.pydata.org/pandas-docs/stable/io.html)
> sep : str, defaults to ',' for read_csv(), \t for read_table()
>     Delimiter to use. If sep is None, the C engine cannot automatically detect the separator, but the Python parsing engine can, meaning the latter will be used automatically. 

...it seemed like there wasn't a huge difference between the two functions.

vreuter:
I'm OK with either way, but `read_table` with explicit `sep=None` may be a tad cleaner/more succinct. 
If we do go with extension, I think that the check should be csv-first and account for case, then fallback to tab. Something like:
```
if extension.lower() == ".csv":
    <read-csv-call>
else:
    <tab-based fallback call>
```

vreuter:
`read_table` errors could be related to default row indexing behavior...`index_col=False` may do the trick if so.

nsheff:
I think it's worth trying to figure out how to do it automatically so we don't have to reinvent the wheel here.

Also if relying on extension (which I think is a bad idea)  should we throw an error if an incorrect extension is given, then?

vreuter:
Agree, I think `read_table` with `sep=None` is best. With extension route, I don't have a strong preference between raising our own exception vs. letting `read_table` do its own thing. If extension-checking, I think the first case is CSV, second is [TSV/TXT] (all case-agnostic), then exception.

ninjha01:
I'm having trouble getting `read_table` to guess file types.

`pd.read_table(sample_file, sep=",", engine='python', index_col=False)` 

works without a hitch on csv's, but 

`pd.read_table(sample_file, sep=None, engine='python', index_col=False)`

fails with the following error:

`ValueError: Annotation sheet ('/tmp/pytest-of-nj7kv/pytest-25/test_minimal_configuration_out0/anns.csv') is missing column(s): sample_name`

@nsheff @vreuter any ideas?

vreuter:
Can you push whatever changes you have so that I can run the tests and see the full traceback?

nsheff:
I don't nkow how pandas read_table works. if this doesnt work we could aslo try `csv.Sniffer` [see here](https://stackoverflow.com/questions/16312104/can-i-import-a-csv-file-and-automatically-infer-the-delimiter).

vreuter:
The tests :heart: the changes now; I'm good to merge this; what do you think @nsheff ?

nsheff:
I'm getting errors:
` The 'dtype' option is not supported with the 'python' engine`

vreuter:
OK I can check it out

PR #116, opened 2017-06-02
Remove resources requirement in PipelineInterface

vreuter:
This should solve https://github.com/epigen/looper/issues/28

### DIFF
--- a/looper/models.py
+++ b/looper/models.py
@@ -70,6 +70,7 @@ from .utils import \
     parse_ftype, check_bam, check_fastq, get_file_size, partition
 
 
+DEFAULT_COMPUTE_RESOURCES_NAME = "default"
 DATA_SOURCE_COLNAME = "data_source"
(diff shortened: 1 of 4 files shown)


looper/models.py
```diff
@@ -1126,24 +1127,40 @@ def as_data_frame(self):
         return _pd.DataFrame([s.as_series() for s in self.samples])
 
 
-    def to_csv(self, path):
+    def write(self, path, sep=None):
```
nsheff: should we keep a `to_csv` as a wrapper of `write` for backwards compatibility?

vreuter: I'd prefer to have just `write`, but only slightly. It's totally up to you; I'll add it back if you'd like.

vreuter: Agree, good call

nsheff: I will go ahead and merge this, we can add back in if need be.

vreuter: Sounds good.

Issue #96, opened 2017-05-05
implied columns to supplant genome and transcriptome sections


nsheff:
I just implemented a new feature called `implied_columns`. This is now active in the `dev` branch and I've added documentation, which you can read here:

http://looper.readthedocs.io/en/dev/implied-columns.html

Implied columns is a generic solution to how we're currently handling `genome` and `transcriptome`. I propose we deprecate these specific sections (which are not documented anywhere anyway) and use the generic `implied_columns` concept.

Looking for feedback from @afrendeiro, @jklughammer, and anyone else who has interest.

nsheff:
See also #94 for code.

vreuter:
This seems to be working well, can this be closed @nsheff ?

nsheff:
The implied columns is great, yeah -- but I guess the remaining question is: can we eliminate the old sections?

Maybe just stick a warning in there if someone specifies a `genomes` or `transcriptomes` section in project_config ?

vreuter:
So you want a requirement for `genomes` or `transcriptomes` to imply a requirement to use `implied_columns`?

nsheff:
No I just mean we deprecate the old use of `genomes`, and provide a warning if someone still uses it, with a suggestion to switch to implied columns.

vreuter:
Ah :ok_hand: sounds good

afrendeiro:
Fine with me on eliminating genomes/transcriptomes from config.

PR #122, opened 2017-06-06
Finish merging protocol mapping and pipeline interface

### LINKED ISSUE #55: Proposal: merge pipeline interface and protocol mapping into one config file
To prevent relying on a hard-coded folder structure, let's merge protocol mappings into a component of pipeline_interface.yaml and move those up one level, and then make the paths to pipelines in protocol_mappings relative to this file; 

this eliminates the hard-coded paths of 'config' and 'pipelines' subfolders we are currently employing.

### LINKED ISSUE #88: increase verbosity of required inputs not found
Small thing: if a pipeline is not submitted because input files are not found, right now looper says:
```
 Not submitted: Required input files not found
```
I think in this case it should state specifically which files are missing. It does in `--debug` mode but this might be worth displaying to the user.

nsheff:
* Fixes #88
* Move toward #96 
* Fix #55

### DIFF
--- a/looper/looper.py
+++ b/looper/looper.py
@@ -253,23 +253,18 @@ def run(prj, args, remaining_args, interface_manager):
         # cannot be assigned (library/protocol missing).
         for pipeline_interface, pipeline_key, pipeline_job in pipelines:
 
+            # pipeline_key (previously pl_id) is no longer necessarily script name, it's more flexible.
             # The current sample is active.
             # For each pipeline submission consideration, start fresh.
             skip_reasons = []
 
-            # Discard any arguments to get just the (complete) script name,
-            # which is the key in the pipeline interface.
-            pl_id = os.path.basename(str(pipeline_job).split(" ")[0])
-            # pl_id is no longer necessarily script name, it's more flexible.
-            pl_id = pipeline_key
-
             _LOGGER.debug("Setting pipeline attributes for job '{}' (PL_ID: '{}')".
-                          format(pipeline_job, pl_id))
+                          format(pipeline_job, pipeline_key))
 
             try:
                 # Add pipeline-specific attributes.
                 sample.set_pipeline_attributes(
-                        pipeline_interface, pipeline_name=pl_id)
+                        pipeline_interface, pipeline_name=pipeline_key)
             except AttributeError:
                 # TODO: inform about WHICH missing attribute(s).
                 fail_message = "Pipeline required attribute(s) missing."
@@ -288,7 +283,7 @@ def run(prj, args, remaining_args, interface_manager):
 
             # Identify cluster resources required for this submission.
             submit_settings = pipeline_interface.choose_resource_package(
-                    pl_id, sample.input_file_size)
+                    pipeline_key, sample.input_file_size)
 
             # Reset the partition if it was specified on the command-line.
             try:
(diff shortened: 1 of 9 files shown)


looper/looper.py
```diff
@@ -251,25 +251,20 @@ def run(prj, args, remaining_args, interface_manager):
         # Go through all pipelines to submit for this protocol.
         # Note: control flow doesn't reach this point if variable "pipelines"
         # cannot be assigned (library/protocol missing).
-        for pipeline_interface, pipeline_job in pipelines:
+        for pipeline_interface, pipeline_key, pipeline_job in pipelines:
```
vreuter: nice value-agnostic name

looper/models.py
```diff
@@ -2081,12 +2089,47 @@ def __init__(self, pipedir):
                 _LOGGER.error(str(iface))
                 raise e
 
+    def pipeline_key_to_path(self, pipeline_key):
```
vreuter: It looks like `script_path` and `script_cmd` are swapped here?

vreuter:
Looks great, I just think there may be a variable name swap in one of the functions.

looper/models.py
```diff
@@ -2091,18 +2091,21 @@ def __init__(self, pipedir):
...
-        :param str pipeline_key: the key in the pipeline interface yaml file used
+        :param str pipeline_key: the key in the pipeline interface file used
             for the protocol_mappings section. Previously was the script name.
+        :return (str, str): more restrictive version of input key, along with
+            absolute path for pipeline script.
 
         """
-        # key may contain extra command-line flags; split out to strict key / flags
-        strict_pipeline_key, tmp, pipeline_key_args = pipeline_key.partition(' ')
+        # key may contain extra command-line flags; discard flags.
```
nsheff: we don't discard flags; we retain them because we use them later

looper/models.py
nsheff: where?

looper/models.py
```diff
@@ -1,28 +1,26 @@
...
 
 .. code-block:: python
 
     from models import Project
     prj = Project("config.yaml")
-    prj.add_sample_sheet()
     # that's it!
 
-Explore!
+Explore:
```
nsheff: I got tired of it.

looper/looper.py
nsheff: great!

looper/models.py
vreuter: I mean from the full command, to derive the key. But yes, it's too strong.

nsheff:
I don't think they were swapped, but I renamed them so they are more clear now, is that better?

looper/models.py
```diff
@@ -2018,16 +2023,11 @@ def build_pipelines(self, protocol_name, priority=True):
...
-                             for path in script_paths])
-
-        if priority and len(jobs) > 1:
-            return jobs[0]
+                jobs.append([(ifproto.interface, ) +
+                             ifproto.pipeline_key_to_path(pipeline_key)
+                             for pipeline_key in pipeline_keys])
 
-        return list(itertools.chain(*jobs))
+        return jobs[0] if priority and len(jobs) > 1 else itertools.chain(*jobs)
```
nsheff: ?
```
  File "/home/ns5bc/.local/bin/looper", line 9, in <module>
    load_entry_point('looper==0.6.0.dev0', 'console_scripts', 'looper')()
  File "/home/ns5bc/.local/lib/python2.7/site-packages/looper/looper.py", line 816, in main
    run(prj, args, remaining_args, interface_manager=interface_manager)
  File "/home/ns5bc/.local/lib/python2.7/site-packages/looper/looper.py", line 235, in run
    if len(pipelines) == 0:
TypeError: object of type 'itertools.chain' has no len()
```

vreuter:
Should also address https://github.com/epigen/looper/pull/122

PR #123, opened 2017-06-06
Release version 0.6

nsheff:
Proposed release update. See changelog for details.


afrendeiro:
Sorry, I'm getting an error by simply starting a Project object:

```python
In [2]: from looper.models import Project
In [3]: prj = Project("metadata/project_config.yaml")
AttributeError                            Traceback (most recent call last)
<ipython-input-3-76f87f954b56> in <module>()
----> 1 prj = Project("metadata/project_config.yaml")

/home/arendeiro/.local/lib/python2.7/site-packages/looper/models.pyc in __init__(self, config_file, subproject, default_compute, dry, permissive, file_checks, compute_env_file, no_environment_exception, no_compute_exception)
    405         if subproject:
    406             _LOGGER.info("Using subproject: '{}'".format(subproject))
--> 407         self.parse_config_file(subproject)
    408 
    409         self.name = self.infer_name(self.config_file)

/home/arendeiro/.local/lib/python2.7/site-packages/looper/models.pyc in parse_config_file(self, subproject)
    669             # Relative to environment config file.
    670             self.compute.submission_template = _os.path.join(
--> 671                     _os.path.dirname(self.environment_file),
    672                     self.compute.submission_template
    673             )

/cm/shared/apps/python/2.7.12/lib/python2.7/posixpath.pyc in dirname(p)
    120 def dirname(p):
    121     """Returns the directory component of a pathname"""
--> 122     i = p.rfind('/') + 1
    123     head = p[:i]
    124     if head and head != '/'*len(head):

AttributeError: 'NoneType' object has no attribute 'rfind'
```
So without passing any compute config, it seems looper isn't filling those with the defaults correctly for me.

It's hard for me to see when should those be filled although I suspect the issue is already with the `.compute` attribute still being None. One comment on that: I noticed the usage of something like `if variable` statement when parsing the arguments of a function. I would suggest trading that with `if variable is not None` to make it explicit and handle the various cases (e.g. I could pass `False` to that kwarg thinking it has another behaviour, or pass a list which also evaluates as False).

afrendeiro:
I'm getting an error with both looper run, check and summarize:

```bash
arendeiro@n002 arid1a$ looper summarize metadata/project_config.yaml 
Command: summarize (Looper version: 0.6.0-dev)
Creating Project from file: 'metadata/project_config.yaml'
Loading LOOPERENV: /home/arendeiro/.local/lib/python2.7/site-packages/looper/submit_templates/default_looperenv.yaml
Using default LOOPERENV. You may set environment variable LOOPERENV to configure environment settings.
Parsing Project config file
Paths section in project config is deprecated. Please move all paths attributes to metadata section. This option will be removed in future versions.
  File "/home/arendeiro/.local/bin/looper", line 11, in <module>
    load_entry_point('looper==0.6.0.dev0', 'console_scripts', 'looper')()
  File "/home/arendeiro/.local/lib/python2.7/site-packages/looper/looper.py", line 791, in main
    compute_env_file=getattr(args, 'env', None))
  File "/home/arendeiro/.local/lib/python2.7/site-packages/looper/loodels.py", line 40, in __init__
    no_compute_exception=RuntimeError, **kwargs)
  File "/home/arendeiro/.local/lib/python2.7/site-packages/looper/models.py", line 431, in __init__
    self.add_sample_sheet()
  File "/home/arendeiro/.local/lib/python2.7/site-packages/looper/models.py", line 914, in add_sample_sheet
    row_dict = merge_rows.iloc[row].to_dict()
  File "/home/arendeiro/.local/lib/python2.7/site-packages/pandas/core/indexing.py", line 1328, in __getitem__
    return self._getitem_axis(key, axis=0)
  File "/home/arendeiro/.local/lib/python2.7/site-packages/pandas/core/indexing.py", line 1749, in _getitem_axis
    self._is_valid_integer(key, axis)
  File "/home/arendeiro/.local/lib/python2.7/site-packages/pandas/core/indexing.py", line 1638, in _is_valid_integer
    raise IndexError("single positional indexer is out-of-bounds")
IndexError: single positional indexer is out-of-bounds
```
I believe this has to do with the usage of `.iloc` (I have pandas 0.20.1).
`iloc` and `irow` both require integers as arguments (or boolean arrays) and maybe for some reason the index is not (maybe related with the [new pandas version](http://pandas.pydata.org/pandas-docs/version/0.20/whatsnew.html#v0-20-1-may-5-2017) or with the switch to `read_table`)? I can't seem to reproduce this when I read a table in the same way as looper does however.
An alternative syntax would be to use `.loc[row, :]`

nsheff:
I can reproduce the first error. I hadn't tried instantiating a project like this yet, sorry.

But I cannot reproduce the second error. I'm using pandas 0.20.2... can you also post the results of running it with `looper --dbg` ?

nsheff:
(or try upgrading pandas to 0.20.2 and see if that fixes it?)

nsheff:
probably it's because my project is not using a merge table...

vreuter:
@nsheff I'll work on these issues; there must not be a case with a merge table, or at least not one that triggers this.

vreuter:
What was your config file like for reproduction of the `Project` instantiation problem? I can create ratbrain fine with just a config path.

vreuter:
TL;DR -- I think the `Project` creation exception mentioned is expected, but that expectation could be changed. But I'm not sure why I don't encounter it for similar instantitiation attempts, even outside of the `looper` project directory.

For the `Project` creation issue, an exception should be raised when the `Project` class being used is from `models`, but not when it's from `loodels` (or whatever the name becomes). `Project` in `models` should not know about the `looper` default submission templates. This can change--I have no opinion on whether those belong with `looper` specifically or with the what becomes of `models` in `pepkit`--but right now they're intended to be used with the `looper`-specific `Project` rather than the generic one.

nsheff:
So we need to use `from looper.loodels import Project` ?

nsheff:
Yes, that works for me. But importing from looper.models still fails with the crypic error.

nsheff:
>What was your config file like for reproduction of the Project instantiation problem? I can create ratbrain fine with just a config path.

mitodyn for example. microtest works, though...

vreuter:
Yeah `looper` should be importing its `Project` model from `loodels`.

vreuter:
OK, yeah `mitodyn` reproduces for me, too. What do you think -- should generic `models` use the default submission templates, or should those be kept `looper`-specific?

nsheff:
To me, everything in project is still generic (so I actually do not really see the distinction yet). I think there is no looper-specific anything in Project. At least, it's not yet clear to me that these should even be separated. they should simply be optional features of a generic project.

vreuter:
Did you have a final thought on name for compute settings  environment variable, @nsheff ? Some ideas (as alternatives to `COMPUTE_SETTINGS`), which could be underscored (or not) between words/abbreviations: 
`PEPENV`
`PEPCMP`
`PEPCMPENV`
`PRJCMP`
`PRJCMPENV`
`PROJECT_COMPUTE`

nsheff:
I guess I prefer PEPENV, only because it may eventually contain things other than compute...

vreuter:
Sounds good. Succinct, suggests the transition to `pepkit`, indicates that it stores data about an environment, and is flexible like you said. :trophy: :trophy: :trophy: :trophy:

vreuter:
Thanks for catching these @afrendeiro ! Let me know if the new commits solve the issue that you encountered with indexing during the merge table step. I tried out an example that failed in that fashion and that works now, but it'd be good to have this confirmed on the original case. If it fails again, can you share the configuration and merge files you're using, and any other relevant files.

These most recent couple commits allow the generic `Project` to use the default submission templates, which should solve the `Project` creation problem, now whether you're using the general `Project` or the `looper`-specific one, so you can disregard the earlier comments about the difference between those two classes in how they behave when a default isn't explicitly passed.

looper/models.py
```diff
@@ -934,11 +934,27 @@ def add_sample_sheet(self, csv=None):
                             # so we don't re-derive them later.
                             merged_cols = {
                                     key: "" for key in merge_rows.columns}
-                            for row in merge_rows.index:
+                            for row in range(len(merge_rows.index)):
```
afrendeiro: Just one comment on this: while this does fix the issue of passing an integer to `.iloc` the better option would be to keep iterating through the index itself and pass the values to `.loc` which accommodates all types of indexes. As far as I can see that would never raise an error.

vreuter: This was bothering me, too; it feels dirty (at least in `pandas` to use the integer indexing). I stumbled across `iterrows` in the `pandas` docs, which I don't remember using before, but I've put that in here and in one other spot since it sounds built for this purpose and feels most straightforward with what the iteration is doing.

nsheff:
ok is this good to go then? @vreuter I'll merge upon your approval

vreuter:
Yep -- 3, 2, 1... :rocket:

PR #135, opened 2017-06-15
0.6 RC2

### LINKED ISSUE #129: Extending Sample from library-specific object not working
Hi,

Are there tests cases on the current implementation of sample-type (library)-specific Sample objects that extend Sample [as described in the docs](http://looper.readthedocs.io/en/latest/models.html#extending-sample-objects)?

This works on 0.5 but is not on 0.6rc1 (only generic Sample objects are created).

### LINKED ISSUE #130: Idea: sample subclasses in pipeline interface
Motivated by #129.

Right now there are 2 protocol-sample mappings: the protocol_mappings file, and then the `__library__` attribute in a sample subclass. This is a problem; if a user changes the protocol mappings, they cannot (easily) change the `__library__` attribute, and so they won't get the appropriate subclass.

Solution: 

We make it so that subclasses need not have a `__library__` attribute, but should just always be used by default. Any class in a `py` pipeline script that extends `Sample` will be used. This solves 95% of use cases, and leaves only 1 problem: What if a pipeline script needs to define two different subclasses for two different protocols?

In this case, the pipeline will need to map these subclasses to protocols NOT through the `__library__` attribute (as it does now), but instead through a new section in the `pipeline_interface` file, like so:

```
pipeline_interface:
  atacseq.py:
    sample_subclasses:
      dnase: DnaseSeqSample
      ATAC-seq: AtacSeqSample
```

This makes it modulated by the pipeline user at the level of the pipeline interface, instead of in the python code itself.

### LINKED ISSUE #132: Fail sooner or with better message if project config points to nonexistent pipeline interface
Right now, if a project config file stores a path to a nonexistent pipeline interface file in its `pipelines_dir` spot, the failure message talks about `protomap` and `ProtocolInterfaces` rather than the nonexistent file. It should be more informative and/or fail sooner.

### LINKED ISSUE #133: Generalize pipelines_dir
The `pipelines_dir` section of project configuration files and the corresponding `Project` attribute has become multiple and polymorphic, allowing pointer(s) to folder(s) and/or file(s). The naming should be generalized.

First ideas:
`pipeline_pointers`
`pipeline_locations`
`pipeline_interfaces`
`pipelines`

This could wait until 1.0 and an immediate hard deprecation (simpler, but harsher) or be part of 0.7 and allow the current format until 1.0 (gentler, but harder).

### LINKED ISSUE #134: Updated requirements
v0.6 introduced a new requirement on pandas 0.20.1 and numexpr... was it 2.26?

* [x] update requirements so installs automatically update pandas and numexpr
* [x] add new requirement to changelog.

### LINKED ISSUE #136: Proposal: eliminate Sample's knowledge of its Project
It feels perhaps too big to tackle in 0.6, or at least unnecessary, but the need for a `Sample` to have knowledge of its `Project` feels slightly strange. Quite possibly, though, there's substantial benefit to this that I don't realize since my primary interaction with `Project` and `Sample` has been one-sided and over a shorter timeframe than others'. So this may ultimately be closed with little or no change, but I wanted to track this thought that's arisen a couple of times as I've been looking at other things.

### LINKED ISSUE #137: TypeError in v0.6-rc1
reported by @rcorces

In python 3, we get an error : `TypeError: unorderable types: sttr() > int()`
from looper.py line 353: `if submit_settings.setdefault("cores", 1) > 1:`

argument seems to be getting passed as a string.

fixing it like so works:
```
if int(submit_settings.setdefault("cores", 1)) > 1:
                    cmd += " -P " + submit_settings["cores"]
                try:
                    if int(submit_settings["mem"]) > 1:
                        cmd += " -M " + submit_settings["mem"]
```

### LINKED ISSUE #139: pipeline interface path variable should accept environment variables
Most paths can use environment variables, like `$HOME`, but when I tried it in the `pipeline_interface : pipelines : path` it choked. We should run this through `expandpath`, too.

vreuter:
Close #129 
Close #130 
Close #133 
Close #136 
Close #137 
Close #139 
Close #134 
Close #132 

- Update requirements.
- Separate some logical concerns within `Project` constructor and `Sample` subclass seeking.
- Clean up messaging about sample run skip reasons.
- Remove outdated example files.
- Require `setuptools` that's now generally assumed for install
- Faster and more informative failure for nonexistent `pipelines_dir` interface file


nsheff:
I confirm that this solves the missing interface problem in my test.

nsheff:
But this did *not* force a new reliance on pandas 0.20 when I had pandas 0.19 installed, so the requirement thing is not working yet.

nsheff:
My commits fix #134 for me. Previously, an old version of pandas was satisfying it because the version number was getting stripped off when we read in the requirements file. @vreuter please double-check  	f58fc74

nsheff:
Addresses #133

vreuter:
When you get the chance @afrendeiro can you rerun the case that caused #129 with `looper --dbg ...` and see what the debugging output indicates regarding what's happening as the `Project` searches for the `Sample` subclasses?

afrendeiro:
Okay, I've pulled the latest `0.6-rc2` and ran looper with --dbg.
A few issues with this:
1. for a successful install `setup.py` needs to stop looking for the script files that were removed;
1. why `argparse.SUPPRESS` on the log options? Good for simplicity but prevents non-devs from seeing the logs...

Here's a sample from a `looper --dbg check` (highlighted important bits):

```rst
looper:787 [INFO] > Command: check (Looper version: 0.6.0-rc2) 
...
models:632 [WARNING] > Looper v0.6 suggests switching from pipelines_dir to  pipeline_interfaces. See docs for details. 
...
models:905 [DEBUG] > Adding sample sheet 
models:915 [DEBUG] > Creating samples from annotation sheet 
models:1162 [DEBUG] > **Successfully imported pipelines **
models:999 [DEBUG] > merge: **sample** 'ATAC-seq_HAP1_ACTB_r1_2471_10'; lane=3 
models:999 [DEBUG] > merge: sample 'ATAC-seq_HAP1_ACTB_r1_2471_10'; data_source=/scratch/lab_bsf/samples/BSF_0279_HG7MMBBXX/BSF_0279_HG7MMBBXX_3_sample
...
```
So looper is still finding my pipeline_dirs but making generic `Sample` objects. I don't understand.
Do you guys have any pipeline with `__library__` to test this?

One more issue with this update: instantiating a `Project` no longer calls `add_sample_sheet` internally.
Wasn't the consensus on this that they would be tied? For `looper` CLI usage there's no difference since looper will call that but for interactive use this again implies a change...

nsheff:
do you have a module called 'pipelines' installed, and does this module have a matching sample subclass? This is where it will be looking (not in the pipeline itself), by my reading...

nsheff:
yeah, the current logic is not the way I thought it worked. this is going to take some reworking I think. It's not doing what it used to somehow...

@afrendeiro: do you want it to be pulling the child class from a global 'pipelines' install (bad way imo because it relies on the magic name 'pipelines'), or do you want it to pull child class from the python file that defines the pipeline (preferable in my opinion).

Right now it's doing the former and not the latter as far as I can tell; I think we should switch to the latter and exclude the former.

vreuter:
I deliberately pulled two functions from the tail of the `Project` constructor with https://github.com/epigen/looper/commit/49f1e5f138f3a67d7f528025d6b276b5746b5f45, because I was unaware of previous discussion about this. In general, I much prefer to separate things, but if there's consensus / I'm outvoted to include the calls at the end of the constructor logic, that's OK.

I see two main disadvantages--only one of which has unknown/potentially significant extent, the other is known and bounded--and one advantage, of inclusion in the constructor. The significant disadvantage is that this complicates inheriting `Project`. If a potential client developer wanted to derive from `Project` but have additional constructor logic before those calls at the end, he or she would need to modify the constructor itself rather than just overriding the functions that would be called at its end / post-construction. While probably not a frequent use case, this feels like a more substantial issue. The minor one is that this complicates testability, either requiring mocking or eroding test logic independence. The advantage is as mentioned, that it reduces needed function calls when using a `Project` in isolation, outside of the CLI.

vreuter:
Regarding the `Sample` subclasses, the logic could also get messed up if the import search finds a valid `pipelines` before a globally installed package. This sometimes happens when within the directory from which the `import` is attempted there's a *module* (or another package) with the same name as the targeted, globally-installed package.

looper/models.py
```diff
@@ -441,6 +441,8 @@ def __init__(self, config_file, subproject=None,
         self.sheet = None
         self.samples = list()
 
+        self.add_sample_sheet()
```
vreuter: Did this *not* break tests?

vreuter: Interestingly, it didn't -- sweet! `add_sample_sheet` has idempotence, at least in the context of the test cases where the `Project` fixture that calls this is used.

nsheff: Yes, it is indeed idempotent (universally actually). I knew this before, and had at one point just left it in both places...

we can go back to the other way if you want. I was the proponent for moving them in, since I couldn't imagine a situation where you'd want to create a project and NOT add the sample sheet.

vreuter: Yeah me either, so I'm ok with it back in too. Just worried about the subclassing thing which seems like a potential pain but like I said probably quite uncommon

vreuter: And even then it'd only matter if the client developer needed additional code to execute *before* adding the sample sheet

nsheff: why not add an option in init to shut off the auto-populate functions? then you can isolate things if you need to but the default is to just run everything through on the main init.

vreuter:
Changed tilte since this isn't yet stable

vreuter:
@nsheff do you have some other cases you normally put through dry run to check how things are looking? `microtest` is well-behaved on the newest commit and is picking up ATAC-specific sample with a slightly tweaked pipeline interface file, but it'd be good to have some additional checks.

vreuter:
Also, what about the `SUPPRESS` on the log options @nsheff ? I don't have a strong preference either way. @afrendeiro do you think it's better to display them? I think they were included at one point and were later suppressed.

nsheff:
personally I would say don't clutter the CLI with the log options, but let's add them to the readthedocs in an advanced section.

nsheff:
got this error:

```
looper run -d examples/test_project/test_config.yaml 
Command: run (Looper version: 0.6.0-rc2)
Loading PEPENV: /home/nsheff/.local/lib/python2.7/site-packages/looper/submit_templates/default_compute_settings.yaml
Using default PEPENV. You may set environment variable PEPENV to configure environment settings.
Looper v0.6 suggests switching from pipelines_dir to pipeline_interfaces. See docs for details: http://looper.readthedocs.io/en/latest/
  File "/home/nsheff/.local/bin/looper", line 11, in <module>
    load_entry_point('looper==0.6.0rc2', 'console_scripts', 'looper')()
  File "/home/nsheff/.local/lib/python2.7/site-packages/looper/looper.py", line 819, in main
    compute_env_file=getattr(args, 'env', None))
  File "/home/nsheff/.local/lib/python2.7/site-packages/looper/loodels.py", line 29, in __init__
    no_compute_exception=RuntimeError, **kwargs)
  File "/home/nsheff/.local/lib/python2.7/site-packages/looper/models.py", line 630, in __init__
    else self._make_basic_samples()
  File "/home/nsheff/.local/lib/python2.7/site-packages/looper/models.py", line 1009, in _make_basic_samples
    if _os.path.isfile(self.metadata.merge_table):
  File "/usr/lib/python2.7/genericpath.py", line 37, in isfile
    st = os.stat(path)
TypeError: coercing to Unicode: need string or buffer, NoneType found

```

nsheff:
random suggestion:
>  20 of 20 sample(s) generated job template(s); 17 of 23 job template(s) were considered for submission, and 17 of those were actually submitted.

Let's get rid of the repeated "(s)" and either make the computer infer (it knows the #) or just use s always. it looks a bit ridiculous like that :)

actually let's go a bit further and just make this a list instead of a sentence, So it says:

```
Samples generating jobs: 20/20
Jobs considered for submissIon: 17/23
Jobs submitted: 17/17
```

nsheff:
also for me on microtest sample-yaml is not getting passed:

```
/home/nsheff/code//pipelines/pipelines/atacseq.py --sample-yaml None -G hg19 -I ATAC-seq_human_PE  -O /home/nsheff/microtest/results_pipeline -P 4 -M 16000  | tee /home/nsheff/microtest/submission/ATAC-seq_human_PE_atacseq.log
```

vreuter:
Can you give this another run whenever you have the chance @nsheff ? `microtest` seems to be giving me the expected behavior now, both in the case of when `pipelines` doesn't supply a specific subtype and when it does. With regard to the YAML file for the submission script, that is.

nsheff:
oops, the `examples` folder got removed somehow; these are used in the docs

```
source/sample-annotation-sheet.rst:27: SEVERE: Problems with "csv-table" directive path:
[Errno 2] No such file or directory: '../examples/microtest_sample_annotation.csv'.
```

vreuter:
That removal was intended because the examples are non-functional. I think that we should either include them but have them be functional, or exclude them and change the docs.

nsheff:
yeah. I think they should be included because it makes maintaining examples in the docs easier. So, we need to make them functional. actually this is a perfect example -- we must keep them functional otherwise they shouldn't be included in the docs. they should probably also be incorporated into our tests -- tested docs! nice! how's that for keeping docs up-to-date...

nsheff:
actually now after thinking for a few minutes my excitement has faded a little... they don't need to be *really* functional, just *potentially* functional, to be a useful example. And, I think, they are...

vreuter:
OK, agreed. Question, though--will the docs break if it's made more apparent in the examples that they aren't intended to be functional (i.e., by making the paths & such things like `/path/to/output_dir`)? I ask because recently when I tried to run one of these rather than the real `microtest` I had to pause and look at the actual config file to see that it was an issue with path existence rather than something internal to `looper`. I can imagine something similar happening for an external user. He or she may not initially look at the config file's content, or think that `looper` intends to pick up the CeMM-specific paths.

nsheff:
Nope. for the docs they just need to *exist*. the docs are easy to please. so yes, they can become obviously non-functional examples...

nsheff:
@vreuter How close are we to merging this?

vreuter:
I'm set with it now. Did you have other cases you wanted to run?

nsheff:
I'll give it a try with some of the ones I'm working on now. I'm mostly just itching to move on

vreuter:
Are you making more changes?

vreuter:
One note on the `Sample` subclasses that I noticed during this round of changes...it seems like some of the `Sample` subclass constructors assume that the input data is a `Series`. This became apparent when I'd represented a generic `Sample` as an ordinary `dict` and then passed that data through to the subtype constructor(s) within `looper.run`. This works fine, having `Series` be the base type that's passed, but in general I think it may make more sense to have the `Sample`subclass constructors also operate on a normal dictionary and/or an `Iterable` of two-tuples since they're more basic/builtin types. This could lower the barrier to someone using a `Sample`, eliminating a need for familiarity with `pandas`.

PR #149, opened 2017-07-10
Interactive logging, explicit option names, subtype docs & tests

### LINKED ISSUE #131: Feature request: interactive logging
Can we have a way to turn on the logger with a single function, when using interactive projects?

Motivated by #129

### LINKED ISSUE #146: Explicit CLI option names
From @afrendeiro 
> [really minor thing: can the argparse arguments that receive values (e.g. --logfile ) have a dest= keyword making explicit the destination variable? This makes it easy to track where certain variables come from/are used again.]

Benefits:
1. Homogeneity with options that present a different name to the user than is used internally
2. Concordance with explicit > implicit idea.

### LINKED ISSUE #147: Document Sample subclass declaration and use
From @afrendeiro 

> How is this implemented exactly now? This section of the docs probably needs to be updated. I couldn't figure out how to let looper know about my classes.
Importantly, this should be independent of looper and just handled by models. Not entirely sure if it is like that right now the flow now is build_submission_bundles -> fetch_sample_subtype -> _import_sample_subtype but build_submission_bundles is only called by looper and not somewhere at Project initialization as I'd have thought. In fact there is no logic in Project to accommodate sample subclasses, which now only uses _make_basic_samples.
But despite working, this means that Sample subclasses are only triggered by looper and are not present in the samples of the Project. This would be really useful for the future use of models outside of looper! (and my current use too) - can we please have this implemented in this way?

vreuter:
This validates `setup_looper_logger` as a way to establish logging and provide it for an interactive session. It also adopts explicit names for CLI options as discussed in #146 .
Close #131 
Close #146 
Close #147


nsheff:
can you add some short instructions for how to use the interactive feature in the docs?

vreuter:
Yep definitely

vreuter:
Is there an existing docs page that you have in mind? Maybe `tutorials.rst` or `usage-and-commands.rst`, or something new?

nsheff:
This is going to be docs for models, not for looper (which won't really be used interactively); so:

http://looper.readthedocs.io/en/latest/models.html

vreuter:
Got it, sounds good.

vreuter:
As I was writing the docs, it occurred to me that it may not actually be necessary to define the subtype in the same module as the pipeline, just that it's available (i.e., imported) in the pipeline. There are now some tests confirming that, but I think it's best to not be directly advertising that in the docs just yet.

PR #151, opened 2017-07-13
0.6 rc3

### LINKED ISSUE #131: Feature request: interactive logging

### LINKED ISSUE #138: Display subcommand description in its own help text
With `looper --help`, I can see the description for each available CLI program. While not critical, it'd be nice to also have this description with `looper <subcommand> --help`.

### LINKED ISSUE #144: v0.6rc2 improvements/clarifications
Hi Vince and Nathan,
I had a go with v0.6rc2 and since I encountered several hiccups (now in retrospect I think mostly caused by old docs), I just to be sure none of the issues was caused by me and I just took some time to go over them.
My comments follow:
- Logging:
    - Could there be an easy way to let the user see some of the logging options in the CLI without cluttering too much the `--help` message? Right now one has to look at the code to activate more verbose logging.
    - ~~Similarly, could we add Vince's way of logging in interactive section (which I somehow cannot find anywhere in the pull/issue discussions) to the docs?~~ #131 <- already an issue
    - the `--dbg` option doesn't seem to do anything for me, I still get the same amount of detail. When using the interactive session however (as described in #129), I do get more detail.
    - [really minor thing: can the argparse arguments that receive values (e.g. [--logfile](https://github.com/epigen/looper/blob/master/looper/looper.py#L59) ) have a `dest=` keyword making explicit the destination variable? This makes it easy to track where certain variables come from/are used again.]
- Sample subclasses <small>(this took me quite a while to understand - some of the comments actually don't apply, but since I wrote them as I explored, they reflect my steps in trying to understanding how this works and can be useful to make a few things more explicit/documented)</small>:
    - How is this implemented exactly now? [This section of the docs](http://looper.readthedocs.io/en/latest/models.html#extending-sample-objects) probably needs to be updated. I couldn't figure out how to let looper know about my classes.
    - Importantly, this should be independent of looper and just handled by `models`. Not entirely sure if it is like that right now the flow now is `build_submission_bundles` -> `fetch_sample_subtype` -> `_import_sample_subtype` but `build_submission_bundles` is only called by looper and not somewhere at Project initialization as I'd have thought. In fact there is no logic in Project to accommodate sample subclasses, which now only uses `_make_basic_samples`.
    - I got some confirmation of the above when I saw the following message in my looper log: "'/home/arendeiro/workspace/pipelines/pipelines/rnaBitSeq.py' appears to attempt to run on import; does it lack a conditional on '__main__'? Using base type: Sample" <- so it is trying to use subclasses! (This pipeline in particular does not use classes and does not extend Sample though). When I inspected my Yaml files of another data type/pipeline, my additional attributes were there! Maybe the logger should be more refined here is saying something like "Using subclass X for sample A" or maybe just my logging from CLI doesn't work.
    - But despite working, this means that Sample subclasses are only triggered by looper and are not present in the samples of the Project. This would be really useful for the future use of models outside of looper! (and my current use too) - can we please have this implemented in this way?
- Sample-Project interaction/link:
    - Can we reverse the removal of Sample-Project linking? It is true these don't need to exist, and probably shouldn't be used in the looper workflow but are extremely useful outside of it.

TODO:
- [x] Logging CLI help (#145)
- [x] Explicit CLI option names (#146)
- [x] Document Sample subclasses (#147)
- [x] Logging options docs (#148)
- [x] relink samples/projects (for now -- to be revisited later) (#150)

### LINKED ISSUE #145: Logging info in CLI help
From @afrendeiro 

> Could there be an easy way to let the user see some of the logging options in the CLI without cluttering too much the --help message? Right now one has to look at the code to activate more verbose logging...
the `--dbg` option doesn't seem to do anything for me, I still get the same amount of detail. When using the interactive session however (as described in #129), I do get more detail.

In the information provided, describe the requirement on the location of `--dbg` within the command in order to get the desired effect.

### LINKED ISSUE #146: Explicit CLI option names

### LINKED ISSUE #147: Document Sample subclass declaration and use

### LINKED ISSUE #148: Advanced logging options documentation and CLI subcommand
@nsheff replying to @afrendeiro 

> Logging:
Could there be an easy way to let the user see some of the logging options in the CLI without cluttering too much the --help message? Right now one has to look at the code to activate more verbose logging.

> I propose: a detailed documentation page in readthedocs, with a note in the CLI help message.

vreuter:
Main differences between rc2 and this version include:
1 -- more thorough documentation, particularly for logging options available via CLI, the extension of `Sample` for the creation of subtypes, and interacting with `models`
2 -- partial restoration of bidirectionality of link between a `Sample` and the `Project` that created it
3 -- clever scripted updates to usage docs

Minor differences:
1 -- explicit names for looper CLI options in code via `dest` `argparse` keyword arg
2 -- confirmation of availability of interactive session logging
3 -- `looper` subcommand description in its own help text.

Close #148 
Close #147 
Close #146 
Close #145 
Close #144 
Close #138 
Close #131


vreuter:
Did I miss a commit when I deleted `dev`, renamed my merge branch to `dev`, then pushed? :worried:

nsheff:
I don't think so -- I was updating the autodocs to incorporate your changes to the usage.

vreuter:
OK, whew! It felt a little dirty to delete `dev` and then swap in a renamed branch, but I checked the `git log` pretty carefully

nsheff:
actually now that I look there's something weird bout the template, not sure where it got messed up but I'll correct it

nsheff:
ok done!

PR #152, opened 2017-07-17
User path expansion and cleaner cluster submission

vreuter:


vreuter:
Closing in favor of intra-fork version

PR #153, opened 2017-07-17
Paths and cluster submission

vreuter:


nsheff:
can you summarize (in 1-2 bullets) the improvement here (for changelog)? just coding style, no user improvements?

vreuter:
Sure, I think there are 4 main things:
1. I ran into an issue specifying a `pipelines_dir` value with a `~`; it looks like we'd only been calling `expandvars`; `expanduser` allows the `~`-prefixed specification.
2. Functional concern separation, as you mentioned.
3. Checking that the path to a pipeline actually exists before writing the submission script. Specifically, to prevent something from being submitted (and dying immediately) for an unavailable script. I think it's nice to handle this with `looper` since it's something that's being processed by the logic we have in place, and it seems simpler for the user to look at a helpful message from `looper` than to go hunting for a cluster manager log.
4. Support `path` value in a pipeline interface that's just a folder. So if I have a key like `ATACseq.py`, I can just write `path: ${CODE}/ATACseq/pipelines` rather than  `path: ${CODE}/ATACseq/pipelines/ATACseq.py`

vreuter:
Sorry, not exactly two bullets there but wanted to enumerate each unit...

PR #158, opened 2017-07-24
Resolve failing tests

vreuter:
The failures in the build of current `master` stem from lack of warning message, generation of which is recovered here.


nsheff:
Ok let's just sneak this into master to fix the failing build...

PR #166, opened 2017-08-09
add required arg to file path call

### LINKED ISSUE #165: Mismatch between docs example and required usage pattern
`Sample`/`Project` separation changed the signature for `Sample.set_file_paths` in such a way that invalidates one of the examples in the docs (i.e., a call in the example code is missing a required argument). The implementation may revert to a signature that lacks the additional parameter/required argument, but it may make sense to change in the `latest` docs version?
This is visible in the `Logistics` section of https://github.com/epigen/looper/blob/master/doc/source/models.rst
See discussion in https://github.com/epigen/open_pipelines/issues/2

vreuter:
Fix #165


vreuter:
Now contained within https://github.com/epigen/looper/pull/167

PR #178, opened 2017-10-11
Only check flags for config samples

vreuter:
I moved and renamed a couple random other, unrelated things that I noticed, but the main things to look at are the `fetch_flag_files()` function and the `Checker` class
Resolve https://github.com/epigen/looper/issues/177

### DIFF
--- a/looper/looper.py
+++ b/looper/looper.py
@@ -13,7 +13,7 @@ import subprocess
 import sys
 import time
 import pandas as _pd
-from . import setup_looper_logger, LOGGING_LEVEL, __version__
+from . import setup_looper_logger, FLAGS, LOGGING_LEVEL, __version__
 from .loodels import Project
 from .models import \
     grab_project_data, ProjectContext, Sample, \
@@ -130,6 +130,14 @@ def parse_arguments():
     destroy_subparser = add_subparser("destroy")
     check_subparser = add_subparser("check")
     clean_subparser = add_subparser("clean")
+
+    check_subparser.add_argument(
+            "-A", "--all-folders", action="store_true",
+            help="Check status for all project's output folders, not just "
+                 "those for samples specified in the config file used")
+    check_subparser.add_argument(
+            "-F", "--flags", nargs='*',
+            help="Check on only these flags/status values.")
 
     # Common arguments
     for subparser in [run_subparser, summarize_subparser,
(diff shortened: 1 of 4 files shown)


looper/looper.py
```diff
@@ -877,35 +877,54 @@ def uniqify(seq):
...
-    # TODO: resume here to hook into specific protocols, and if we want
-    # TODO (continued): to just use Python rather than shelling out.
-    from collections import defaultdict
-    flags_by_protocol = defaultdict(list)
+    def _glob_expr(flag):
+        return os.path.join(
+            prj.metadata.results_subdir, "*", "{}.flag".format(flag))
+
+    # TODO: import from pep.
+    flags = ["completed", "running", "failed", "waiting", "partial"]
```
nsheff: the `partial` flag is new to me here

nsheff: I think we should also move the flag list out to a global variable, we may use it elsewhere.

looper/looper.py
```diff
@@ -877,35 +877,54 @@ def uniqify(seq):
...
-    counts = {}
-    for f in flags:
-        counts[f] = int(subprocess.check_output(
-                pf + "*/*" + f + ".flag 2> /dev/null | wc -l", shell=True))
 
-    for f, count in counts.items():
-        if 0 < count < 30:
-            _LOGGER.info(f + " (" + str(count) + ")")
-            subprocess.call(pf + "*/*" + f + ".flag 2> /dev/null", shell=True)
+class Checker(Executor):
```
nsheff: is there an advantage in this case to making this a function object instead of just a function?

nsheff:
I think my review somehow got posted on the wrong commit, but anyway, a couple of first thoughts:
 - do you really think we need a function class here instead of just a function for checking?
- what do you think of @rcorces' suggestion to have a `-a` option to preserve current behavior, is that unnecessary?
- what is the new `partial` flag?
- I think we should move the flag list to a global setting somewhere

vreuter:
Cool thanks for the review. I'll try to match up replies to each item:
- The `Checker` class is just to match what's already in place with the other `looper` programs now (`Runner`, `Cleaner`, etc.). I think the framework for it is here: https://github.com/epigen/looper/commit/15298d6a76dddae566d26fb836301bc04b09eba5
- Ah yeah forgot about the `-a`; I think it's a good idea and will add it in. I actually first did that way, then converted, so it won't be too tricky to have both.
- "partial" is there for use with the checkpoint concept in `pypiper`
- Totally agree, the flags should be at higher scope. I was going to do that along with pep migration but it could be done now also, within `looper`. I bubbled them up in `pypiper` so wouldn't hurt to do the same here.

nsheff:
Ok sounds good. On the flags -- would it be possible to make this function totally open so it could handle any possible flags? What if other software uses different flags, etc?

vreuter:
Yes

vreuter:
OK, capability for just Project samples and all folders within output dir is in now. I went with `-A` and `--all-folders` for the short/long options to trigger for `check`. There's also the ability to just check for specific flag(s) now with `-F/ --flags`

PR #209, opened 2018-03-31
Ensure path expansion for data sources

vreuter:
When running the `rnapipe` example, it looks like the data file paths are declared for the structure where the files are relative to the config. We need to ensure that paths are made absolute in this case, which this does, under the assumption that a relative declaration is with respect to the config file location. This also changes the timing of the env var expansion, doing it when the project parsing and creation is done rather than when sample files are located, so that we're not repeating that process unnecessarily. Not a major deal, but brings actual code into better alignment with what's going on conceptually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ open_pipelines/
 # pytest-related
 .cache/
 .coverage
+.pytest_cache/
 
 # Reserved files for comparison
 *RESERVE*
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 ******************************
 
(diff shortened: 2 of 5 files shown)


nsheff:
I guess otherwise it was relative to the location of the `pwd` ? this worked previously for me, but only when I ran it from the correct directory. so, this changes that to make it relative to the config file and not to the `pwd` right? am I thinking about it correctly?

vreuter:
Yeah, that's the intention. I try to run from a project folder but often don't, so I guess I'm a good test case for behavior when that's not the case ;D

vreuter:
Yeah I've added a couple little notes now, but obviously feel free to modify as you see fit. I think you have a better feel for the docs than I do.

PR #214, opened 2018-06-15
use orderedDict for sample. Fix #213

### LINKED ISSUE #213: Samples have the wrong attribute ordering
samples have  `as_series` and `get_sheet_dict` functions, which return Ordered lists of attributes -- but the orders of the attributes are incorrect. They are not maintaining the order of the attributes in the file (so, for example, `sample_name` ends up in the middle of the OrderedDict, which causes problems for `looper summarize`, which produces strangely ordered column tables.

nsheff:
@vreuter 
This solves the problem for me. I can now install peppy and it fixes the downstream looper error that was making unordered stats sheets. There are 2 little things I thought might be really easy for you to help me with, that are taking me awhile, if you don't mind...

1. I have tried to add a test. I can create a simple one:
```
class SampleOrderTests:

    def test_sample_order():

	import conftest
	import peppy

        proj = conftest.interactive()
        s = peppy.Sample(proj.sheet.iloc[0])
        d = s.get_sheet_dict()
        assert list(d)[0] == 'sample_name'
```

This makes sure the `sample_name` is the first column. a simple unit test. But I cannot figure out how to insert this into the complex pytest framework.

2.  With my change, the tests succeed in python 2, but they fail in python 3 with this error: 

`TypeError: to_dict() takes 1 positional argument but 2 were given` when I try to run `pytest`.

do you know off the top of your head how to accomplish this in python 3?

### DIFF
--- a/peppy/sample.py
+++ b/peppy/sample.py
@@ -55,7 +55,8 @@ class Sample(AttributeDict):
     def __init__(self, series, prj=None):
 
         # Create data, handling library/protocol.
-        data = dict(series)
+        data = OrderedDict(series)
+        _LOGGER.debug(data)
         try:
             protocol = data.pop("library")
         except KeyError:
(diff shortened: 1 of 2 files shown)


vreuter:
This works for me on 2.7.10 and 3.6.5, both with `pyhon setup.py test` and `pytest`. Well see what Travis says but let me know if you test this out locally and how it goes.

nsheff:
strange that it worked on py3 for you but not locally for me...

PR #217, opened 2018-06-27
release 0.18.0

nsheff:


doc/source/changelog.rst
```diff
@@ -1,5 +1,22 @@
...
+
+    - Fixed a bug that caused an install error with newer ``numexpr`` versions.
+
+  - New
+
+    - Project names are now inferred with the ``infer_name`` function, which uses a priority lookup to infer the project name: First, the ``name`` attribute in the ``yaml`` file; otherwise, the containing folder unless it is ``metadata``, in which case, it's the parent of that folder.
+
+    - Add ``get_sample`` and ``get_samples`` functions to ``Project`` objects.
+
+    - Add ``get_subsamples``and ``get_subsample`` functions to both ``Project`` and ``Sample`` objects.
```
vreuter: Can the function names and class name more closely match? Either `Subsample` and the current function names, or the current class name and the `get_sub_samples` and `get_sub_sample`?

doc/source/changelog.rst
```diff
@@ -1,5 +1,22 @@
...
+
+    - Subsamples are now objects that can be retrieved individually by name, with the ``subsample_name`` as the index column header.
```
vreuter: Again, I like "Sub" as part of the name of the entity, but then I think the class name should be `Subsample` rather than `SubSample`

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
...
+        project_name = os.path.basename(config_folder)
+        
+        if project_name == "metadata":
+            project_name = os.path.basename(os.path.dirname(config_folder))
+
+        self.name = project_name
+
+        return self.name
+
+    def get_subsample(self, sample_name, subsample_name):
```
vreuter: Would prefer to keep with the (mostly?) consistent double-spacing between member function definitions.

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
...
+
+        s = self.get_sample(sample_name)
+        return s.get_subsample(subsample_name)
+
```
vreuter: Same here on spacing

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
...
+
+    def get_samples(self, sample_names):
+        """
+        Returns a list of sample objects given a list of sample names
 
-        :param str path_config_file: path to the project's configuration file.
-        :return str: name of the configuration file's folder, to name project.
+        :param list sample_names: A list of sample names to retrieve
+        :return: List: A list of Sample objects
```
vreuter: `List` should be `list`, and ideally list[Sample]. Also there's an extra colon between the tag and the type

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
         return os.path.join(os.path.dirname(__file__), "submit_templates")
 
 
-    @staticmethod
-    def infer_name(path_config_file):
+    def infer_name(self):
```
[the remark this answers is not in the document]
vreuter: Actually, I see now that this is called in the constructor, so the name makes more sense I think...

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
...
+        if hasattr(self, "name"):
+            return(self.name)
+        
+        config_folder = os.path.dirname(self.config_file)
+        project_name = os.path.basename(config_folder)
+        
+        if project_name == "metadata":
+            project_name = os.path.basename(os.path.dirname(config_folder))
+
+        self.name = project_name
```
vreuter: Actually, why is the effect moving into this function rather living out in the constructor?

peppy/project.py
```diff
@@ -441,20 +441,67 @@ def templates_folder(self):
         return os.path.join(os.path.dirname(__file__), "submit_templates")
 
 
-    @staticmethod
```
vreuter: If this is going to be used for effect and no longer be static, the logic should live directly in the constructor rather than in this function.

peppy/sample.py
```diff
@@ -28,6 +28,19 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+@copy
+class SubSample(AttributeDict):
+    """
+    Class to model SubSamples
```
vreuter: This should have a more complete doc

tests/models/independent/test_Project.py
```diff
@@ -693,7 +693,9 @@ def test_data_sources_derivation(self, proj, sample_index):
         # We don't care about that here, or about duplicates.
         expected = set(DERIVED_COLNAMES)
         observed = set(merged_columns)
-        assert expected == observed
+        # Observed may include additional things (like auto-added subsample_name)
+        for val in expected:
```
vreuter: Could be helpful to diagnose all failures in one shot with something like `assert set() == expected - observed`

peppy/project.py
nsheff: yeah that's funky. no idea how that got through...

peppy/project.py
nsheff: it isn't moving... it was already in the function...

peppy/project.py
nsheff: does it seem more natural to override a function in a child class?

peppy/project.py
vreuter: It was in the constructor, but `infer_name` itself didn't set anything, at least not in the current `master` version, right? It looked like the value from `infer_name` was used by the constructor to set `name`

peppy/project.py
vreuter: Ahhhh I see! Hadn't thought about overrides; that definitely makes more sense if the idea is to encourage or at least facilitate overriding.

peppy/project.py
nsheff: Ah, now I see what you mean. Ok, I can go with that way. 

I think I did it this way because to me, the function name `infer_name` for some reason implied a side-effect. that felt more logical. but to you it doesn't -- so I will move the setting logic back into the constructor.

peppy/project.py
nsheff: yeah -- originally you were overriding this in the `Project` child in `looper`.

peppy/sample.py
```diff
@@ -371,6 +385,37 @@ def library(self):
...
+        Retrieve a single subsample by name.
+
+        :param str subsample_name: The name of the desired subsample. Should 
+            match the subsample_name column in the subannotation sheet.
+        :return SubSample: Requested SubSample object
+        """
+        subsamples = self.get_subsamples(subsample_name)
+
+        if len(subsamples) > 1:
+            _LOGGER.warn("More than one subsamples was detected; returning the first")
```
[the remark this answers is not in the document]
nsheff: Yeah, this is meant mostly as an experimental feature at this point... I think we haven't really formalized what should happen in the event of multiple samples...

should it throw an error, or return all samples of that name?


I guess `get_subsamples` returns all samples, but `get_subsample` throws an error, since it should always return exactly 1 sample? I don't want it sometimes returning a sample and sometimes a list, since they're different types.

tests/interactive.py
```diff
@@ -7,4 +7,36 @@
 import conftest
 
 print("Establishing Project for testing and exploration")
-proj, pi = conftest.interactive()
+proj = conftest.interactive()
+
+
+import peppy
+import os
+reload(peppy)
```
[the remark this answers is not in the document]
nsheff: this is just my throwaway interactive script that was helping me debug and write functions... when I made changes I was using reload to test them out and it got stuck in there.

doc/source/changelog.rst
nsheff: done.

doc/source/changelog.rst
nsheff: Got it.

peppy/project.py
nsheff: yeah, my bad.

peppy/sample.py
nsheff: got it.

peppy/sample.py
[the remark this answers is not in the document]
nsheff: I implemented an error for duplicate types on attempting to get a subsample. you can always get both with `get_subsamples` if you need to.

peppy/sample.py
```diff
@@ -933,8 +978,18 @@ def merge_sample(sample, sample_subann, data_sources=None, derived_columns=None)
     # Keep track of merged cols,
     # so we don't re-derive them later.
     merged_attrs = {key: "" for key in this_sample_rows.columns}
-
-    for _, row in this_sample_rows.iterrows():
+    subsamples = []
+    _LOGGER.debug(this_sample_rows)
+    subsample_count = 0
+    for subsample_row_id, row in this_sample_rows.iterrows():
+        try:
```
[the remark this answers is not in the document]
nsheff: yeah.

tests/models/independent/test_Project.py
nsheff: true. I guess I find the iterative way a little more readable though

peppy/sample.py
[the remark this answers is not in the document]
nsheff: Ah, but: `AttributeError: 'Series' object has no attribute 'setdefault' `

nsheff:
Ok, how's this?

tests/interactive.py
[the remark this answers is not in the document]
vreuter: OK got it

peppy/sample.py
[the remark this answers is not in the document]
vreuter: Ah OK didn't realize this is/can be `Series` rather than a builtin mapping type

peppy/project.py
vreuter: Ha actually, I can see the side-effecting implication of `infer_name`. It would seem odd to call this from outside the class, though, both semantically and conceptually if it's already in the constructor. So either way--in the constructor or as a separate, non-static function as it is now--seems fine, but if it's as a separate function and the name stays the same, can it be made `_infer_name`?

nsheff: the only time I could see wanting to call this externally would be if you update something (like move it) and want to re-infer the name. Or, if you overrode this function you might want to call it. hence I have a slight preference for keeping it as `infer_name`, but if you think `_infer_name` is better I will change it. I think there is a chance someone would call it externally in some situation.

vreuter: Ah, no I'd missed that the attribute had been entirely removed from this function--feels even better now! Again, GH graphics team should provide the fuzzy blanket icon. I can't be the only one who'd use it.

Issue #229, opened 2018-11-19
activate subproject function


nsheff:
`peppy` needs a function to activate a subproject.

This will allow us to load a `pep`, inspect its subprojects, and then choose one, instead of needing to know them *a priori*.

Issue #228, opened 2018-11-19
sample_annotation should not be required


nsheff:
If you try to load a project without a `sample_annotation` section, you get:

```
_MissingMetadataException: Project configuration lacks required metadata section sample_annotation; used config file '/home/ns5bc/code/aml_e3999/metadata/project_config.yaml
```

But what if I want to load a Project, and then activate a subproject, in two steps. Maybe the initial project need not be complete.

Shouldn't peppy just work anyway in this situation instead of failing?

nsheff:
Related to #224 

The file can be missing/nonexistant, or the attribute can be missing.

Issue #224, opened 2018-11-06
allow missing annotation file


nsheff:
right now if you give a config file that points to a sample annotation file that does not exist, `peppy` does not complete:

```
IOError: [Errno 2] No such file or directory:
```

This should be downgraded to a warning (maybe you want to activate a different subproject)

PR #232, opened 2018-11-27
Subproj related

michalstolarczyk:
addressed issues: #224, #228, #229

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -51,6 +51,7 @@ from collections import Counter
 import logging
 import os
 import sys
+import warnings
 if sys.version_info < (3, 3):
     from collections import Iterable, Mapping
 else:
@@ -234,7 +235,7 @@ class Project(AttributeDict):
             _LOGGER.info("Using subproject: '{}'".format(subproject))
         self.parse_config_file(subproject)
 
-        if "data_sources" in self:
+        if "data_sources" in self and self["data_sources"] is not None:
             # Expand paths now, so that it's not done for every sample.
             for src_key, src_val in self.data_sources.items():
                 src_val = os.path.expandvars(src_val)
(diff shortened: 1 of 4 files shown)


peppy/project.py
```diff
@@ -512,6 +498,17 @@ def get_sample(self, sample_name):
 
         return samples[0]
 
+    def activate_subproject(self,subproject):
```
vreuter: This will overwrite existing entries, but `activate_subproject` _feels_ to me like an operation that would be strictly additive rather than mutative. I don't have a strong preference, but what about `use_subproject`?

michalstolarczyk: we named it "activate" in `pepr`. But I'm ok with changing it

peppy/project.py
```diff
@@ -512,6 +498,19 @@ def get_sample(self, sample_name):
...
+        Activate a subproject.
+
+        This method will update Project attributes, adding new values
+        associated with the subproject indicated, and in case of collision with
+        an existing key/attribute the subproject's value will be favored.
+
+        :param str subproject: A string with a subproject name to be activated
+        :return Project: A Project with the selected subproject activated
+        """
+        return self.__init__(self.config_file, subproject=subproject)
```
vreuter: This should work, some attributes on a `Project` instance that aren't included in a config file that may have been dynamically set by a user post-construction would be reset by this. That may be the intended behavior...what do you guys think? If we wanted the update to be strictly based on the subproject, what about making this `self.parse_config_file(subproject)`?

vreuter: Also, this will return `None`. Is the goal to use a returned value, or just realize the effects of the call? If we just want the effects, we should remove `return` and just make this `self.__init__...` or `self.parse_config_file...`

michalstolarczyk: I didn’t even think about retaining the user-generated changes as I perceived subproject as a fresh start directly from the (slightly changed) config file. 
But you guys have broader experience with PEP usage in the real-life analyses, so maybe it is indeed beneficial to retain the changes.

peppy/project.py
vreuter: Sounds good, since the intent is to update the existing Project rather than create a new one I think that makes more sense than my suggestion

michalstolarczyk:
I can't get the version after updates to activate the `noFrog` subproject in [`example_subprojects2`](https://github.com/pepkit/example_peps/tree/master/example_subprojects2).

Here's what I did:

On Rivanna

```
In [1]: import peppy

In [2]: import inspect

In [3]: lines = inspect.getsource(peppy.Project.activate_subproject)
   ...: print(lines)
   ...: 
    def activate_subproject(self, subproject):
        """
        Activate a subproject.

        This method will update Project attributes, adding new values
        associated with the subproject indicated, and in case of collision with
        an existing key/attribute the subproject's value will be favored.

        :param str subproject: A string with a subproject name to be activated
        :return Project: A Project with the selected subproject activated
        """
        self.parse_config_file(subproject)
        return self
```
So the new version of `peppy` is used
```
In [4]: sp=peppy.Project("/home/mjs5kd/code/example_peps/example_subprojects2/project_config.yaml")

In [5]: sp.subprojects
Out[5]: {'noFrog': {'metadata': {'sample_annotation': 'sample_annotation_noFrog.csv'}}}

In [6]: sp1=sp.activate_subproject("noFrog")

In [7]: sp1.sheet
Out[7]: 
  sample_name organism time                        file_path
0     frog_0h     frog    0   data/lab/project/frog_0h.fastq
1     frog_1h     frog    1   data/lab/project/frog_1h.fastq
2    human_1h    human    1  data/lab/project/human_1h.fastq
3    human_0h    human    0  data/lab/project/human_0h.fastq
4    mouse_1h    mouse    1  data/lab/project/mouse_1h.fastq
5    mouse_0h    mouse    0  data/lab/project/mouse_1h.fastq

In [8]: sp.samples[0]
Out[8]: {'paths': Paths object., 'name': 'frog_0h', 'organism': 'frog', 'sample_name': 'frog_0h', 'required_paths': None, 'derived_cols_done': [], 'time': '0', 'sheet_attributes': ['sample_name', 'organism', 'time', 'file_path'], 'yaml_file': None, 'results_subdir': '/home/mjs5kd/hello_looper_results/results_pipeline', 'file_path': 'data/lab/project/frog_0h.fastq', 'merged': False}

```

The `samples[0]` is still the `frog_0h` which should be deleted as defined in the subproject

vreuter:
Thanks for the detailed example @michalstolarczyk I'll address this tonight

michalstolarczyk:
> Thanks for the detailed example @michalstolarczyk I'll address this tonight

Can confirm that it works now in that case

vreuter:
@michalstolarczyk @nsheff are we ready to merge this now?

michalstolarczyk:
yes

PR #238, opened 2019-01-14
version 0.19

nsheff:
this seems like it's worth a point release
- [x] update changelog
- [x]  bump version


nsheff:
This is actually probably good enough for `0.19` since we've addressed evrything on that milestone: https://github.com/pepkit/peppy/milestone/9

Issue #251, opened 2019-02-18
During subproject activation, preserve/merge data


vreuter:
https://github.com/pepkit/peppy/issues/250

michalstolarczyk:
probably related to: https://github.com/pepkit/caravel/issues/65

vreuter:
See transferred (finally!) https://github.com/pepkit/peppy/issues/250

michalstolarczyk:
Is this supposed to be fixed as well (with #252)? Here's what I get on `dev`:

```python
In [6]: p=peppy.Project("/Users/mstolarczyk/Uczelnia/UVA/code/caravel/examples/hello_lo
   ...: oper/project/project_config.yaml")
No local config file was provided
Found global config file in DIVCFG: /Users/mstolarczyk/Uczelnia/UVA/code/pepenv/uva_rivanna.yaml
Loading divvy config file: /Users/mstolarczyk/Uczelnia/UVA/code/pepenv/uva_rivanna.yaml
Use 'compute_packages' instead of 'compute'
Available packages: set(['singularity_local', 'default', 'largemem', 'singularity_slurm', 'sigterm', 'local', 'parallel'])
Activating compute package 'default'

In [7]: p.metadata.keys()
Out[7]: 
['submission_subdir',
 'pipeline_interfaces',
 'sample_annotation',
 'results_subdir',
 'output_dir']

In [8]: ps=p.activate_subproject("dog")
No local config file was provided
Found global config file in DIVCFG: /Users/mstolarczyk/Uczelnia/UVA/code/pepenv/uva_rivanna.yaml
Loading divvy config file: /Users/mstolarczyk/Uczelnia/UVA/code/pepenv/uva_rivanna.yaml
Use 'compute_packages' instead of 'compute'
Available packages: set(['singularity_local', 'default', 'largemem', 'singularity_slurm', 'sigterm', 'local', 'parallel'])
Activating compute package 'default'

In [9]: p.metadata.keys()
Out[9]: 
['submission_subdir',
 'pipeline_interfaces',
 'sample_annotation',
 'results_subdir']
```

`p.metadata.sample_annotation` is updated, but `p.metadata.output_dir` is missing after subproject activation.

vreuter:
In general, yes, but things like this are likely to continue coming up initially, as a nontrivial number of attributes have been treated specially in the past.

vreuter:
[...]

PR #252, opened 2019-02-18
Fix subproject activation

vreuter:
#251 
Includes also some random styling/docs improvements and a type check on absolute path assurance

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -52,9 +52,9 @@ import logging
 import os
 import sys
 if sys.version_info < (3, 3):
-    from collections import Iterable, Mapping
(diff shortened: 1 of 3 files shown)


nsheff:
what exactly happened that caused this? I thought it would be something in the move to attmap, but was it not?

vreuter:
Hm, no your initial thought seems right. Probably because of the way to handle collisions on a key when the value is a Mapping. In general, the values shouldn't merge (expected behavior is overwrite), so they don’t in attmap. Here, though, when activating a subproject we want the merge behavior, as our implementation of the "priority-configuration" sort of concept, like with command-line vs. config file vs. env var preference.

PR #263, opened 2019-02-25
only set subproject during constructor or subproject activation; https…

vreuter:
…://github.com/pepkit/peppy/commit/3fd3b5f0fc1fe5e30466fdbc3c85a33681db0b38#r32463283


vreuter:
OK thanks!

Issue #264, opened 2019-02-25
Deactivating subproject not documented


michalstolarczyk:
When the `activate_subproject` method is called with `None` as the argument the subproject is _deactivated_. 

I think we should document that or, better yet, use this in a separate `deactivate_subproject` method.

michalstolarczyk:
resolved with https://github.com/pepkit/peppy/pull/265

PR #265, opened 2019-02-26
Deactivate subproject

michalstolarczyk:
Addresses #264

TO DO:

- [x] write corresponding unit tests
- [x] need to resolve: https://github.com/pepkit/attmap/issues/9

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -490,7 +490,8 @@ class Project(AttMap):
         """
         if self.subproject is None:
             _LOGGER.warning("No subproject has been activated.")
-        return self.activate_subproject(None)
+        self.__init__(self.config_file)
+        return self
 
     def activate_subproject(self, subproject):
         """
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -503,6 +503,9 @@ class Project(AttMap):
         :param str subproject: A string with a subproject name to be activated
         :return peppy.Project: Updated Project instance
         """
+        if subproject is None:
+            raise TypeError("The subproject argument can not be NoneType."
+                            " To deactivate a subproject use the deactivate_subproject method.")
         previous = [(k, v) for k, v in self.items() if not k.startswith("_")]
         conf_file = self.config_file
         self.__init__(conf_file, subproject)
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -480,6 +480,18 @@ class Project(AttMap):
         except IndexError:
             raise ValueError("Project has no sample named {}.".format(sample_name))
 
+    def deactivate_subproject(self):
+        """
+        Bring the original project settings back
+
+        This method will bring the original project settings back after the subproject activation.
+
+        :return: peppy.Project: Updated Project instance
+        """
+        if self.subproject is None:
+            _LOGGER.warning("No subproject has been activated.")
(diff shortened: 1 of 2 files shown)


nsheff:
I would instead favor something like `reset_subproject` or something... `deactivate_subproject` makes it seem like I would pass it a string, and it would deactivate *that* specific subproject...  `reset` sounds generic.

michalstolarczyk:
I liked the symmetry activate-deactivate and that's what I would look like as a user. But you're right, "reset" is more reflective of what happens

nsheff:
I actually think I prefer the way it was... just `activate_subproject` with `None` means reset. I think that makes the most sense to me... activating no subproject is the same.

but I'm fine with adding a `reset` function if it makes sense to you. I don't think `deactivate` makes sense to me (that's a completely different concept which would have to retain any user changes, which we are not doing), despite the symmetry.

michalstolarczyk:
I see. I'll make the changes after @vreuter expresses his opinion.

vreuter:
I don't have a strong preference.

I don't really like the mutability we provide with `activate_subproject`; I'd prefer `Project` to be an immutable type entirely fixed by a config file, with subproject "activation" returning a new instance rather than modifying the existing one and returning that. That's totally separate and I think we have no plans for it since I think it's not a broadly shared preference, but it underlies the weakness of preference for how to handle the activation/deactivation naming and arguments.

For discussion, though -- in general I like when significant behavioral differences between functions are implemented (and communicated to users) through truly separate functions rather than by allowing the particular argument to a parameter to correspond to such different behavior. I think there are two benefits:
- A user doesn't need to read past the top level of the API (the object names) to get a good sense of available functionality. With an argument dictating much about a function's behavior, a user needs to invest more time up front in reading documentation to even clearly picture the extent of what's provided.
- There's less risk of unintended behavior. In this specific case, for example, what if someone was calling some function to return the name of a subproject to activate, and that function could return `None`? That user would be at risk of inadvertently deactivating a subproject when the intent was to activate one. If we implement _deactivation_ with the `None`-to-`activate` strategy, we sacrifice an opportunity to provide a fail-fast implementation of _activation_. If we use separate functions, we can forbid `None` as an argument to `activate_subproject` to provide extra usage safety.

vreuter:
I sort of agree that `deactivate` sounds sort of strange for what's going on, but the symmetry is really nice, at least when initially acquainting with the API; I think it helps with that early navigation. Definitely no strong preference on that, though; it feels like the preference may vary a lot among users.
I'm going to annoy myself here and submit a single anecdote as evidence -- and so am doing so tepidly! :stuck_out_tongue_closed_eyes: -- but there _is_ precedent for something similar...
with `conda` there's `conda activate <env>` and `conda deactivate`

michalstolarczyk:
Yeah, the points you mention as bullets are a huge advantage of having a separate method for that and that's why I favored new function over documenting that behavior in #265

vreuter:
Yeah so this looks good to me @michalstolarczyk ; do you want me to add tests, or will you do that?

michalstolarczyk:
I’ll take care of the tests

nsheff:
ok sounds good. I still think 'deactivate' and 'deactivate_subproject' are very different things, but happy to be overruled.

PR #274, opened 2019-03-11
Doc tweaks

vreuter:


nsheff:
thanks for your work on the API documentation... I've been thinking we can really do a better job making this more approachable to newbies.

PR #289, opened 2019-04-29
Snakemake compatibility

### LINKED ISSUE #286: Snakemake object
From https://github.com/snakemake-workflows/dna-seq-gatk-variant-calling/pull/6

We've done most of the naming issues, but now just to wrap it up, we should make a new object, `SnakeProject(peppy.Project)`.

This object should handle:

- renaming the `sample_name` to sample
- rename `subsample_name` to `unit`
- make sure the id columns have indexes

vreuter:
Close #286 
- `SnakeProject` object
- Regard empty strings in tables as null (for `Snakemake` compatibility)
- Better messaging and modularity


nsheff:
I think that's pretty annoying that the 'coverage' test will give a failure if the % decreased...

PR #319, opened 2019-06-19
0.22.1

vreuter:
- Enforce prohibition on storing full `Project` in `prj` ref on a `Sample`
- Define and export a couple of constants that are relevant for `looper`, particularly for writing the disk representation of a `Sample` and then for conducting the submission of samples.


vreuter:
@nsheff this includes an item from the 0.23 milestone (regarding the `prj` attr on a `Sample`), but I'd like to go ahead and release this as a 0.22.1 since I think there are no breaking changes in the API, and a point release of `looper` to address something reported by @afrendeiro could uses this.

vreuter:
@nsheff for review, all of the relevant changes are concentrated in `sample.py`; the others are specifically in tests, are package metadata, or are organizational and/or syntactic changes.

PR #331, opened 2019-09-19
add __reduce__ to Sample and Project; fix #330

### LINKED ISSUE #330: Objects can't be unpickled
Reincarnation of https://github.com/pepkit/peppy/issues/108

```python
>>> import pickle
>>> from peppy import Sample
>>> import pandas as pd

>>> s = Sample(pd.Series(['asd'], index=['sample_name']))

>>> pickle.dump(s, open("sample.pickle", 'wb'))
>>> pickle.load(open("sample.pickle", 'rb'))
TypeError: __init__() missing 1 required positional argument: 'series'
```

Same thing for `Project`:
```python
TypeError: __init__() missing 1 required positional argument: 'cfg'
```

Maybe these classes need a `__getnewargs__` method? 
 https://docs.python.org/3/library/pickle.html#object.__getnewargs__

afrendeiro:
This should restore the ability to pickle/unpickle Sample and Project objects.

### DIFF
--- a/peppy/sample.py
+++ b/peppy/sample.py
@@ -154,7 +154,6 @@ class Sample(PathExAttMap):
             series = series.to_dict(OrderedDict)
         elif isinstance(series, Sample):
             series = series.as_series().to_dict(OrderedDict)
-        self._series = series
 
         # Keep a list of attributes that came from the sample sheet,
         # so we can create a minimal, ordered representation of the original.
@@ -186,11 +185,11 @@ class Sample(PathExAttMap):
         self.paths = Paths()
 
     # The __reduce__ function provides an interface for
-    # correct object serialization with the pickle pickle.
+    # correct object serialization with the pickle module.
     def __reduce__(self):
         return (
             self.__class__,
-            (self._series,),
+            (self.as_series(),),
             (None, {}),
             iter([]),
             iter({"prj": self.prj}.items())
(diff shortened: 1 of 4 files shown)


afrendeiro:
Should obviously add some tests, but I'm not too familiar with the test suite so maybe you guys can have a look and tell me what you think first.

nsheff:
@vreuter this is beyond me at the moment, can you take a quick look?

vreuter:
> @vreuter this is beyond me at the moment, can you take a quick look?

yeah

peppy/sample.py
```diff
@@ -184,6 +185,17 @@ def __init__(self, series, prj=None):
         # analysis time, and a pipeline author vs. a pipeline user).
         self.paths = Paths()
 
+    # The __reduce__ function provides an interface for
+    # correct object serialization with the pickle pickle.
+    def __reduce__(self):
+        return (
+            self.__class__,
+            (self._series,),
```
vreuter: Is the idea with providing just the original (passed to the constructor) data, rather than any data that's been added since construction, that anything that's not passed in via the series is more dynamic/ephemeral and therefore shouldn't be preserved?

vreuter:
@afrendeiro is the type-specific stuff in the `__reduce__` calls for compliance with what's expected by `pickle` (i.e., does it require certain values to be packed in tuples or iterators)?

afrendeiro:
The tuple returned by __reduce__ is indeed just what pickle expects: https://docs.python.org/3/library/pickle.html#object.__reduce__

afrendeiro:
Also added tests now, although one may want to expand them and also include `Sample.as_series` in the suite.

vreuter:
Thanks @afrendeiro -- @nsheff this looks good

PR #346, opened 2020-04-17
v0.30.0

stolarczyk:
### PEP2.0.0 implementation

Not sure how to go about the version number. According to semver guidelines, we should bump to 1.0.0, since backwards compatibility is broken for (at least): subprojects, implied_attributes

wait for:

- [x] https://github.com/pepkit/ubiquerg/pull/27
- [x] https://github.com/pepkit/example_peps/pull/5

todo:

- [x] version bump (maybe go to `0.30.0` ?)
- [x] update changelog


nsheff:
> according to semver guidelines, we should bump to 1.0.0, since backwards compatibility is broken for 

this is not true if you're < 1.0 which indicates some type of beta state.

nsheff:
ok, I've tested looper successfully with a few projects. I think we are ready to release this.

PR #354, opened 2020-08-03
Include license file in source distribution

synapticarbors:
It would be helpful to include the license file in the source distribution.


stolarczyk:
I've already added the license file to the manifest on `dev`. The PR is open we will release it shortly: https://github.com/pepkit/peppy/pull/353

PR #366, opened 2021-04-14
v0.31.1

stolarczyk:
to do: 
- [x] update changelog
- [x] update version


nsheff:
What do you think about this? This yaml patch caused me hours of troubleshooting to figure this fix out: 
https://github.com/pepkit/peppy/pull/366/commits/9e9ec7c595e4e42386482d81d0d768e232d3c658

I can't remember, was this a recent introduction into this package? Do we really need it? Can we do it without that hack?

stolarczyk:
> What do you think about this? This yaml patch caused me hours of troubleshooting to figure this fix out:

I didn't get that error but since it solved it for you then I guess we're good.

> I can't remember, was this a recent introduction into this package? Do we really need it? Can we do it without that hack?


yes, `load_yaml` is a recent introduction. I added it so that peppy and eido don't need to rely on yacman.

nsheff:
> I didn't get that error but since it solved it for you then I guess we're good.

I'm worried about other stuff. The reason you didn't get the error is that you weren't using both peppy and yacman simultaneously. Using both caused the patch to happen twice, which broke things.

> yes, load_yaml is a recent introduction. I added it so that peppy and eido don't need to rely on yacman.

Ok, that's great -- but does peppy/eido need the ordering? or can then get by without the patch?

stolarczyk:
we used to read the config file in with `yaml` modulde

https://github.com/pepkit/peppy/blob/0d1da63a232509aba112919bb5fa28230c768641/peppy/project.py#L123-L124

which didn't preserve the order. This makes sense since all we need is access to the config attributes. We never write it to file so the order doesn't matter.

I'll simplify the remote YAML loading function then.

PR #367, opened 2021-06-02
v0.32.0

stolarczyk:
todo

- [x] update version
- [x] add tests for new features
- [x] wait for https://github.com/pepkit/attmap/pull/76
- [x] update attmap requirement
- [x] pass tests
- [x] update changelog
- [x] test manually in some real projects
- [x] Make sure it's tested with looper before release


nsheff:
@stolarczyk Just need a quick comment on this -- right now the tests are failing because `peppy` can't import a PEP that's version 2.0.0 -- it says to use a different version:

> E           peppy.exceptions.InvalidConfigFileException: PEP version is invalid: 2.0.0. Please use version 2.1.0 or import an older version of Project from a submodule in this package, e.g. from peppy.pep200 import Project

Shouldn't the same object be used for 2.0.0 and 2.1.0 since they are backwards compatible? I just want to confirm that this is an error in the version checking, and not in the actual PEPs, since that doesn't make sense to me.

stolarczyk:
I don't remember exactly what behavior did we decide on. I agree that it would be great if one object could accommodate both versions. 

In 2.1.0 the Project object can be created based on sample table, no project config is needed.
Maybe there were some related issues that prevented me from implementing it this way.

nsheff:
a separate object for 1.0 vs 2.0 makes sense to me, but not for 2.1 vs 2.0. 

You had this working with all the 2.1 updates, before we decided to increment the spec to 2.1. Are you saying to update the spec you moved that all into a separate object so 2.0 and 2.1.  I don't quite understand.

I think the issue is more likely just that you didn't update the tests -- but look, tests passed in your previous commit (the one prior to the last commit on this branch, which is what broke the tests.

nsheff:
@stolarczyk I'm still not sure what you were intending here so any guidance would be appreciated.

nsheff:
This commit solves most of the tests by not requiring the pep version to be 2.1.0. However, there are 2 tests that still fail because no pep version is included. Can you explain what you're trying to do here? In your tests, you load a version 1 PEP, I guess with a version 2 object, and then it complains that it lacks the version key. But version 1 PEPs aren't supposed to have the version key. So I guess this is just another error with the tests?

nsheff:
There are now 2 duplicates of the project code: in pep200/project.py is, I guess the "old" version -- which handles PEP version 1.0. I am just confused at the naming here; why did you make pep200 the one that handles PEP 1.0? Is it just a naming error and that should be pep100?

The main project in peppy/project.py handles either 2.0 or 2.1 PEPs.

stolarczyk:
Looks good to me. Make sure it's tested with looper before release.

nsheff:
@nleroy917 Are you satisfied that this is working in your use case?

nleroy917:
Yes it looks good to me

nsheff:
getting ready to release...

PR #380, opened 2022-06-01
Implement `__eq__` operator for Project class

nleroy917:
# Implement `__eq__` operator for `Project` class
## Overview
To support the development of [`pepdb`](https://github.com/pepkit/pephub_db) and rigorously test the `from_dict` method, I added an implementation to support the equality operator (`==`) on two `peppy.Project` instances.

The operator just uses the `dict` representation of all the samples in the sample table of the project to ensure all samples are identical across both instances.

## Equality Criteria
For two `Project` instances to be identical, they must pass the following criteria:
1. The number of rows in both sample tables must be identical
2. Each sample must have identical column names and values

For 2, I am utilizing the `to_dict()` representation of each sample to get a set of key-value pairs and compare those to one another. If one sample has an attribute or associated value the other doesn't, then this fails and the `__eq__` method returns `False`.

## Example Usage
```python
p1 = peppy.Project("path/to/cfg1.yaml")
p2 = peppy.Project("path/to/cfg2.yaml")
p1 == p2
>>> False

p1_a = peppy.Project("path/to/cfg1.yaml")
p1_b = peppy.Project("path/to/cfg1.yaml")
p1_a == p1_b
>>> True
```


Khoroshevskyi:
Neither __eq__ function or from_dict are not working. Because few variables weren't initiated, and __eq__ function returns TRUE even if all variables weren't initiated. Variables that weren't initiated are: 

```
_samples_touched: False
st_index: null
sst_index: null
name: null
description: null
_sample_table: Empty DataFrame
Columns: []
Index: []

```

PR #381, opened 2022-06-13
New approach for equality operator

Khoroshevskyi:
I have added function that will convert all project variables to dictionary. It will allow check if all the variables are set and everything works correctly.


nsheff:
how close is this to merging?

Khoroshevskyi:
I will close this pull request, and open new one, because I am mostly done with new function that initiates peppy object.

Issue #382, opened 2022-06-14
PEP versions should be assigned by default if one isn't provided


nleroy917:
`peppy` should provide a default PEP version if one isn't supplied. This would help in two instances:

1. Right now, `peppy` raises an exception if a project is loaded with a configuration file that does not contain a `pep_version` key-value pair.
2. Attempting to access the `pep_version` attribute of a `peppy.Project` instance after installation with just a sample table causes a critical failure.

Assigning a default version (the latest) would gracefully solve both of these problems.

Khoroshevskyi:
fixed!

PR #383, opened 2022-06-14
Default to latest PEP version if none is supplied

nleroy917:
**Overview:**
Currently, `peppy` breaks if a project is instantiated from a sample table only and one attempts to access the `pep_version` attribute. It also will not let you instantiate a project if the config file does not have the `pep_version` key-value pair. These changes just let you instantiate a project without a `pep_version` and also assigns a sample-table-only project to the latest version.

I also added tests to test these cases.

**Issues:**
#382


nleroy917:
Not sure why the tests timed out

nleroy917:
Tests failed again. Looks like it might be an issue with `attmap`. It is saying it can't connect to github.com.

PR #389, opened 2022-07-18
Improve the performance of peppy

### LINKED ISSUE #387: Performance analysis
Peppy works fine for small-ish projects (like, up to thousands of samples), but if you have a huge PEP with tens to hundreds of thousands of samples, it can take prohibitively long to load.

We should do some benchmarking and see if there's any way to improve the performance.

### LINKED ISSUE #388: Optimize peppy code
After #387 there is a possibility to optimize the code so that peppy works better.

rafalstepien:
- merge two similar loops into one
- get rid of accessing instance attribute multiple times

closes #388
closes #387

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -4,12 +4,15 @@ Build a Project object.
 import os
 from collections.abc import Mapping
 from logging import getLogger
+from typing import List, Tuple
 
 import pandas as pd
 from attmap import PathExAttMap
 from pandas.core.common import flatten
 from rich.progress import track
 from ubiquerg import is_url
+
+from peppy.sample import Sample
 
 from .const import (
     ACTIVE_AMENDMENTS_KEY,
@@ -51,8 +54,13 @@ from .const import (
     SUBSAMPLE_TABLE_INDEX_KEY,
     SUBSAMPLE_TABLES_FILE_KEY,
 )
-from .exceptions import *
-from .sample import Sample
+from .exceptions import (
+    IllegalStateException,
+    InvalidConfigFileException,
+    InvalidSampleTableFileException,
+    MissingAmendmentError,
+    SampleTableFileException,
+)
 from .utils import copy, is_cfg_or_anno, load_yaml, make_abs_via_cfg, make_list
 
 _LOGGER = getLogger(PKG_NAME)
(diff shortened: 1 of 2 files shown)


nleroy917:
Your tests might be failing due to the `attmap` dependency in `requirements-dev.txt`. I had this issue as well... Check here: https://github.com/pepkit/peppy/pull/383#issuecomment-1155622589

peppy/project.py
```diff
@@ -473,23 +481,63 @@ def _auto_merge_duplicated_names(self):
...
+        for sample in samples:
+            if sample[st_index] == duplication:
+                duplicated_samples.append(sample)
+            else:
+                not_duplicated_samples.append(sample)
+        return duplicated_samples, not_duplicated_samples
+
+    @staticmethod
+    def _all_values_in_the_list_are_the_same(list_of_values: List) -> bool:
+        return all(value == list_of_values[0] for value in list_of_values) == 1
```
nleroy917, line 539: Does `all()` return a `bool`? Do we need the ` == 1`?. Also, not sure it matters, but reading [here](https://stackoverflow.com/questions/3844801/check-if-all-elements-in-a-list-are-identical), using `itertools.groupby` might be faster? I'm sure you've already read this thread, however.

nleroy917 (approved):

peppy/project.py
```diff
@@ -4,13 +4,16 @@
...
 from logging import getLogger
+from typing import List, Tuple
 
 import pandas as pd
 from attmap import PathExAttMap
 from pandas.core.common import flatten
 from rich.progress import track
 from ubiquerg import is_url
 
+from peppy.sample import Sample
```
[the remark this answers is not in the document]
rafalstepien: AFAIK it is recommended to use absolute imports over relative imports, mostly because of readability (please see eg. [here](https://realpython.com/absolute-vs-relative-python-imports/), or Google style guide [here](https://google.github.io/styleguide/pyguide.html#22-imports))

peppy/project.py
rafalstepien: You are right, we probably don't need `==1`. As to the method used it looks like both of them present similar performance 
image

Issue #386, opened 2022-06-28
Pandas nan object downstream issues


Khoroshevskyi:
By solving #378 and #379 and creating https://github.com/pepkit/pephub_db we found an issue in creating dict with variables from peppy object. Pandas creats `nan` value for every empty column. When we are calling `to_dict` function the `nan` value will be in dictionary that we created and leads to downstream issues.
There is few options to solve this problem: 
1) recursively change all `nan` values to None in the dict.
3) Change `nan` to empty string `""`
2) Change pandas object: https://stackoverflow.com/questions/14162723/replacing-pandas-or-numpy-nan-with-a-none-to-use-with-mysqldb
@nleroy917

Khoroshevskyi:
fixed

Issue #379, opened 2022-05-24
Built-In Method for Comparing two `peppy.Project` classes


nleroy917:
It would be nice to have a reliable way to compare two project classes. At present, a simple equals operator doesn't work:
```python
import peppy
p1 = peppy.Project("example_peps-cfg2/example_basic/project_config.yaml")
p2 = peppy.Project("example_peps-cfg2/example_basic/project_config.yaml")
p1 == p2
>>> False
```

Maybe comparing the samples and ensuring they are identical in each project class?

This would be useful in tests for #378

Khoroshevskyi:
you should compare all the variables in the object, not just dicts, that were generated by `to_dict()`.
Try to convert object variables using this functin: https://github.com/pepkit/pephub_db/blob/648dfcb6d0dc15d8b758cbffc04839ae6cc303fd/peppy_p.py#L15-L48 and then compare it.

nleroy917:
## Meeting on May 31st, 2022:
- Two PEPs are identical if all their samples are identical
- Two samples are identical if each attribute is identical across the two samples

Khoroshevskyi:
fixed

PR #390, opened 2022-07-22
Release `v0.33.0`

nleroy917:
Wait for:
- [x] Update change log
- [x] Change version

## Overview
This pull request encompasses **4 main changes** to `peppy`. Mostly, these changes are to support `pephub` and the new **database representation** of PEPs. Specifically, these changes let us instantiate a PEP from memory (from a `dict`) instead of a config file on disk. The two main issues with reading and writing PEPs "in the cloud" are 1. Instantiating PEPs and retrieving information on PEPs from disk is slow, and 2. PEP storage quickly becomes quite large, especially with large sample files (e.g. from GEO). As such, @Khoroshevskyi has implemented some new methods on the `peppy.Project` object that let us 1. Instantiate a new `peppy.Project` using a `dict` or `json` representation of a PEP, and 2. Compare two PEPs to test for equality, which allows us to validate the in-memory instantiation. There are also minor changes to improve `peppy` performance for large PEPs and better defaults to support and stabilize `pephub`.

## In This Pull Request
**1. Implement `__eq__` operator for Project class:**
The `peppy.Project` class now supports native python equality operations. We've added a `__eq__` method to the class, which lets us compare to project objects like so:

```python
p1 = peppy.Project("path/to/cfg1.yaml")
p2 = peppy.Project("path/to/cfg2.yaml")
p1 == p2
>>> False

p1_a = peppy.Project("path/to/cfg1.yaml")
p1_b = peppy.Project("path/to/cfg1.yaml")
p1_a == p1_b
>>> True
```

**2. Added `from_dict` and `to_dict_extended:`**
A new `from_dict` method was added to the `peppy.Project` class to allow a user to instantiate a new project from a `dict` representation of a PEP. This `dict` might come from some other code, or even a database (in the case of `pephub`). As well, `to_dict` was altered to contain an `extended` flag that creates a richer dictionary representation to support the aforementioned functionality.

**3. Improve the performance of peppy:**
Parsing duplicate samples was found to be redundant and @rafalstepien did some code optimizations to make this fast. Specific changes were:
- merge two similar loops into one
- get rid of accessing instance attribute multiple times

**4. Default to latest PEP version if none is supplied :**
Initially, `peppy` would raise an Exception if no PEP version was supplied. This was causing issues for `pephub` and preventing proper access to valid PEPs. As such, I changed the `peppy.Project` class to warn the user no version was supplied and assume the default PEP version instead of crashing.

## Relevant Issues
#388 
#387 
#386 
#382 
#379 
#378


nleroy917:
@Khoroshevskyi @Khoroshevskyi Feel free to edit my PR notes and let me know if I missed anything!

peppy/project.py
```diff
@@ -140,16 +150,146 @@ def __init__(
...
+            new_dict = {}
+            for key, value in project_value.items():
+                if key != "_project":
+                    new_dict[key] = self._convert_to_dict(value)
+            return new_dict
+
+        elif isinstance(project_value, PathExAttMap):
+            new_dict = PathExAttMap.to_dict(project_value)
+            return self._convert_to_dict(new_dict)
+            # return new_dict
```
rafalstepien, line 191: Comment to delete

peppy/project.py
```diff
@@ -140,16 +150,146 @@ def __init__(
...
+
+        elif isinstance(project_value, pd.DataFrame):
+            project_value = project_value.to_dict()
+            project_value = self._nan_converter(project_value)
+            return project_value
+
+        else:
+            return project_value
+
+    def _nan_converter(self, nan_dict: dict) -> dict or list:
```
rafalstepien, line 205:
```suggestion
from typing import Dict, List, Union
    def _nan_converter(self, nan_dict: Dict) -> Union[Dict, List]:
```

peppy/project.py
```diff
@@ -140,16 +150,146 @@ def __init__(
...
+            for key, value in nan_dict.items():
+                new_dict[key] = self._nan_converter(value)
+            return new_dict
+        elif isinstance(nan_dict, float):
+            if math.isnan(nan_dict):
+                return None
+        else:
+            return nan_dict
+
+    def from_dict(self, d: dict) -> None:
```
rafalstepien, line 228: Maybe replace `d` variable with something more verbose? Eg. `dictionary_representation`, `pep_dictionary_representation`?

peppy/project.py
```diff
@@ -140,16 +150,146 @@ def __init__(
...
+        :param dict d: in-memory dict representation of processed pep.
+        """
+        _LOGGER.info(f"Processing project from dictionary...")
+        if CONFIG_KEY not in self:
+            self[CONFIG_KEY] = PathExAttMap()
+
+        # extract custom index for sample table if exists
+        self.st_index = (
+            d[SAMPLE_TABLE_INDEX_KEY] if SAMPLE_TABLE_INDEX_KEY in d else None
+        )
```
rafalstepien, line 241: Just another example where you can replace a comment with helper function :grinning: :

 ```suggestion
        self.st_index = self.extract_custom_index_for_sample()
        
        (...)
        def extract_custom_index_for_sample(self, pep_dictionary: Dict):
            return pep_dictionary[SAMPLE_TABLE_INDEX_KEY] if SAMPLE_TABLE_INDEX_KEY in pep_dictionary else None
        )
```

File: peppy/project.py (hunk 3 above)
Comment (nleroy917) [reply peppy/project.py:228]: I think I wrote this function, and I agree. I'm partial to `pep_dictionary`

File: docs/changelog.md [hunk 5]
```diff
@@ -2,6 +2,15 @@
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format.
 
+## [0.33.0] -- 2022-06-14
+
+### Changed
+
+- `pep_version` is no longer a required parameter to create a `peppy.Project` instance from a configuration file.
+
```
Khoroshevskyi, line 10: Add new changes here

Khoroshevskyi (approved):

peppy/project.py
nleroy917: I manually implemented this one

Khoroshevskyi (approved):

PR #396, opened 2022-08-17
Adjust peppy to handle different read types for the same sample

rafalstepien:
- Add handler for `AttributeError` raised by `getattr()` and return empty string if there is no value for given attribute
- Add test

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -296,11 +296,11 @@ class Project(PathExAttMap):
             p_dict["_samples"] = [s.to_dict() for s in self.samples]
         return p_dict
 
-    def create_samples(self, modify=False):
+    def create_samples(self, modify: bool = False):
         """
         Populate Project with Sample objects
         """
-        self._samples = self.load_samples()
+        self._samples: List[Sample] = self.load_samples()
         if modify:
             self.modify_samples()
         else:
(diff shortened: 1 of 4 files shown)


nsheff:
I think you will need to lint this, right?

peppy/project.py
```diff
@@ -674,6 +660,43 @@ def _get_duplicated_and_not_duplicated_samples(
...
+            for sample in duplicated_samples:
+                attribute_value_for_sample = Project.safe_getattr(sample, attr)
+                attribute_values.append(attribute_value_for_sample)
+
+            merged_attributes[attr] = list(flatten(attribute_values))
+
+        return merged_attributes
+
+    @staticmethod
+    def safe_getattr(object: object, attribute: str):
```
nsheff, line 694: yes, I like this!

rafalstepien:
> I think you will need to lint this, right?

I think I already did

nsheff (approved):
I didn't try running it, but the code changes look good to me!

peppy/project.py
```diff
@@ -594,56 +594,42 @@ def _auto_merge_duplicated_names(self):
...
-                    disable=not self.is_sample_table_large,
-                )
-                if sample_names_list.count(x) > 1
-            ]
-        )
-        if not dups_set:
-            # all sample names are unique
+        duplicated_sample_ids = self._get_duplicated_sample_ids(sample_names_list)
+
+        if not duplicated_sample_ids:
```
nleroy917, line 599: Maybe, `if len(duplicated_sample_ids) == 0:` ?

Small thing, just slightly better readability.

peppy/project.py
```diff
@@ -1222,6 +1245,27 @@ def _read_sample_data(self):
...
+                "index_col": False,
+                "keep_default_na": False,
+                "na_values": [""],
+            }
+            try:
+                return pd.read_csv(pth, sep=infer_delimiter(pth), **csv_kwargs)
+            except Exception as e:
+                raise SampleTableFileException(
+                    f"Could not read table: {pth}. "
+                    f"Caught exception: {getattr(e, 'message', repr(e))}"
```
nleroy917, line 1266: Is this a typo? Why 3  arguments to `getattr`?

nleroy917: Oh. Is it a default value if the attribute doesn't exist?

peppy/project.py
```diff
@@ -674,6 +660,43 @@ def _get_duplicated_and_not_duplicated_samples(
...
+                attribute_value_for_sample = Project.safe_getattr(sample, attr)
+                attribute_values.append(attribute_value_for_sample)
+
+            merged_attributes[attr] = list(flatten(attribute_values))
+
+        return merged_attributes
+
+    @staticmethod
+    def safe_getattr(object: object, attribute: str):
+        try:
```
nleroy917, line 695: Is this code not equivalent to `getattr(object, attribute, "")`? It might not be, just curious of the differences.

nleroy917 (approved):
Looks good! I just had questions for my own learning...

peppy/project.py
rafalstepien: Oh it is. I didn't know that getattr has default value!

peppy/project.py
rafalstepien: This were changes proposed by linter so I haven't seen that

rafalstepien: But yes, it is a default value

peppy/project.py
rafalstepien: According to PEP8 
image

rafalstepien: https://peps.python.org/pep-0008/

PR #397, opened 2022-08-17
Initializing peppy object from pandas 

Khoroshevskyi:
Added few lines of code that enables initializing peppy.Project from pandas df


peppy/project.py
```diff
@@ -99,12 +104,13 @@ class Project(PathExAttMap):
...
-        sample_table_index=None,
-        subsample_table_index=None,
-        defer_samples_creation=False,
-        project_dict=None,
+        cfg: str = None,
+        amendments: Union[str | Iterable[str]] = None,
+        sample_table_index: Union[str | Iterable[str]] = None,
+        subsample_table_index: Union[str | Iterable[str]] = None,
+        defer_samples_creation: bool = False,
+        project_dict: dict = None,
```
nsheff, line 112: Is it possible have 2 different constructors in Python? Just thinking about this...

or maybe using a helper function to create a Project object from dict, instead of appending these onto the end of the constructor signature ? what do you think?

nsheff (approved):
seems fine to me

nsheff:
Add to changelog

peppy/project.py
```diff
@@ -1,10 +1,12 @@
 """
 Build a Project object.
 """
+from __future__ import annotations
+
```
rafalstepien, line 5: This will not work with Python versions below 3.10. I'm wondering is there a chance that something gets broken?

peppy/project.py
rafalstepien: In the past I used alternative constructor approach in Python with use of classmethods, something like this -> https://code-maven.com/slides/python/class-methods-alternative-constructor. I think it's good idea.

peppy/project.py
```diff
@@ -151,6 +157,11 @@ def __init__(
 
         self.name = self.infer_name()
         self.description = self.get_description()
+
+        if pd_object is not None:
```
rafalstepien, line 161:
```suggestion
        if pd_object:
```

rafalstepien (changes requested):

peppy/project.py
Khoroshevskyi: We can initialize project from dict (dict with all object variables) by calling emty class and than running function from dict.
Now:
`peppy.Project(project_dict="...")
Can be changed to:
`peppy.Project().from_dict(pep_dictionary="...")

Not sure about 2 different constructors...

And in this PR I have added `pd_object` argument

peppy/project.py
Khoroshevskyi: I think we should delete it than!

peppy/project.py
Khoroshevskyi: It won't work! if pd_object will be pandas object, it will through an error

Khoroshevskyi: image

rafalstepien: Ok, weird but that's how pandas works apparently :D

peppy/project.py
Khoroshevskyi: What do you think about that, should we change it?

nsheff: I think I would do it that way.  Keep the constructor clean. SO you might also have a `from_pandas` analogous to `from_dict`, or something, then.

peppy/project.py
nsheff: yes that would be a problem... this will show up in the unit tests, right?

nleroy917: I'm unfamiliar with this. I'm able to import it in python 3.9 on my system:
```console
Python 3.9.13 (main, May 19 2022, 13:48:47) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
>>> from __future__ import annotations
>>> annotations
_Feature((3, 7, 0, 'beta', 1), (3, 10, 0, 'alpha', 0), 16777216)
```

peppy/project.py
nleroy917: More discussion on these things here: https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none

nleroy917 (commented):
I dont have much else to add outside what Rafal and Nathan mentioned!

peppy/project.py
Khoroshevskyi: Ok, I had problems with `@classmethod` , as I think we can't use it there.
Solution: I wrote simple class method that retrieves class. So now it will look like that: 
1) To init class from dict:
`peppy.Project().from_dict(pep_dictionary=___)`
2) To init class from pd object:
`peppy.Project().from_pandas(pandas_df=___)`

peppy/project.py
Khoroshevskyi: There will be problems in python3.10. I have fixed it.

rafalstepien (approved):

PR #403, opened 2022-08-22
400 peppy does not initialize correctly

rafalstepien:
- Fix multiple bugs that poked during initialization of the project with custom index column names
- Add test cases and test data
- Small refactors

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -1,21 +1,20 @@
 """
 Build a Project object.
 """
+import math
 import os
 from collections.abc import Mapping
+from contextlib import suppress
 from logging import getLogger
-from typing import Dict, List, Tuple, Union, Iterable
+from typing import Dict, Iterable, List, Tuple, Union
 
 import pandas as pd
 from attmap import PathExAttMap
 from pandas.core.common import flatten
 from rich.progress import track
 from ubiquerg import is_url
-import math
 
 from peppy.sample import Sample
-
-from .parsers import select_parser
 
 from .const import (
     ACTIVE_AMENDMENTS_KEY,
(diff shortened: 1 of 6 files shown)


peppy/project.py
```diff
@@ -140,13 +139,13 @@ def __init__(
...
         self.sst_index = (
-            subsample_table_index
-            or getattr(self, "sst_index", None)
-            or [
-                SAMPLE_NAME_ATTR,
-                SUBSAMPLE_NAME_ATTR,
-            ]
+            ([subsample_table_index] if subsample_table_index else None)
+            or (
+                [sst_index] if (sst_index := getattr(self, "sst_index", None)) else None
```
nleroy917, line 146: I believe walrus operators are only supported in python >= `3.8`. Just an FYI if peppy promises support for previous versions.

peppy/project.py
```diff
@@ -571,7 +570,7 @@ def _assert_samples_have_names(self):
 
         :raise InvalidSampleTableFileException: if names are not specified
         """
-        try:
+        with suppress(KeyError):
```
nleroy917, line 573: Nice. This is cool, I didn't know it existed.

nleroy917 (changes requested):
Looks good to me! 

I just wanted to raise the question about the walrus operator which I believe is only supported in python 3.8 and above. It seems we advertise that we are supporting down to python v3.6 [here](https://github.com/pepkit/peppy/blob/30e78076c235a4084b944c5cae86048d5a3705ae/setup.py#L46). It's also in our unit tests for the workflows.

Khoroshevskyi (approved):

nleroy917 (approved):

Issue #402, opened 2022-08-19
to_dict() and from_dict() optimization!


Khoroshevskyi:
Right now from_dict returns all possible variables that were initialized in the class. Some of that values are mostly identical.
In the pephub_db it can lead to the huge disc space is taken this variables. It will be a problem in the nearest future. e.g. download 1K projects from GEO.
We need to optimize this functions ASAP.

Issue #401, opened 2022-08-19
from_pandas() function does not initialize '_subsample_df' variable


Khoroshevskyi:
`subsample_df` was not initialized when Project was initializing from_pandas(). 
This issue leads next error:

```
>>> asff = dbc.get_project_by_registry('GEO/GSE146537_samples')
[INFO] [18:40:48] Project has been found: 218
Detecting duplicate sample names ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:--
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bnt4me/.local/lib/python3.8/site-packages/pepagent/pepagent.py", line 261, in get_project_by_registry
    return self.get_project(namespace=namespace, name=name, tag=tag)
  File "/home/bnt4me/.local/lib/python3.8/site-packages/pepagent/pepagent.py", line 294, in get_project
    return peppy.Project().from_dict(project_value)
  File "/home/bnt4me/.local/lib/python3.8/site-packages/peppy/project.py", line 277, in from_dict
    if pep_dictionary[SUBSAMPLE_DF_KEY] is not None:
KeyError: '_subsample_df'
```

PR #405, opened 2022-08-23
changes in the way of initialization Project from dict and pandas df

Khoroshevskyi:
Address #402
Also addresses #401 

Big changes in functions:
- to_dict (returning only crucial object variables. This method is more efficient, than returning all variables)
- from_dict (initialization of the object from dict of variables that were generated by to_dict function)
- from_pandas (added possibility to initialize project also by using subsample table and config file )
- new equal function (Now equal function will check only samples. - previous version was checking all variables. If any path to the file was not the same it would return that this projects were not the same. Now it's fixed)


docs/changelog.md
```diff
@@ -15,6 +15,10 @@ Way of initialization project from dictionary. Now it's possible as follows: Pro
 
 - Feature of initializing `peppy.Project` from `pandas.DataFrame`
 
+## [0.34.0] -- 2022-08-17
```
rafalstepien, line 18: Shouldn't be `0.35.0`? `0.34.0` is the current one

peppy/project.py
```diff
@@ -160,13 +160,10 @@ def __init__(
...
-        dict_self = self._convert_to_dict(self)
-        dict_other = self._convert_to_dict(other)
+        samples_self = [s.to_dict() for s in self.samples]
+        samples_other = [s.to_dict() for s in other.samples]
 
-        dict_self["_samples"] = [s.to_dict() for s in self.samples]
-        dict_other["_samples"] = [s.to_dict() for s in other.samples]
-
-        return dict_self == dict_other
+        return samples_self == samples_other
```
rafalstepien, line 166: You can go direclty for
```python
def __eq__(self, other):
    return [s.to_dict() for s in self.samples] == [s.to_dict() for s in other.samples]
```

tests/smoketests/test_Project.py
```diff
@@ -528,6 +527,7 @@ def test_subtable(self, example_pep_cfg_path):
         Verify that the sample merging takes place
         """
         p = Project(cfg=example_pep_cfg_path)
+        a = p.samples
```
rafalstepien, line 530: Maybe something more verbose than `a` as variable name? :D

Khoroshevskyi: we don't need it. I forgot to delete this line

docs/changelog.md
Khoroshevskyi: I think, this mistake is after merging branches, I'll fix it

rafalstepien (approved):

peppy/project.py
```diff
@@ -356,19 +347,27 @@ def _get_table_from_samples(self, index, initial=False):
         df.set_index(keys=index, drop=False, inplace=True)
         return df
 
-    def parse_config_file(self, cfg_path, amendments=None):
+    def parse_config_file(
+        self,
+        cfg_path: str = None,
+        amendments: Iterable[str] = None,
+        config: dict = None,
```
nsheff, line 354: This should be a separate function, I think.

The function signature like this is not clear. This is similar to what we've discussed before -- this seems like there should be two functions, with different signatures, one that calls the other

nsheff (changes requested):

peppy/project.py
Khoroshevskyi: Thanks for mentioning it! I think, I am even not using this argument! I have to check it again manually

nsheff:
Is the purpose of this to simplify inserting into the database with pepdbagent?

I mean, what prompted the change?

Khoroshevskyi:
> Is the purpose of this to simplify inserting into the database with pepdbagent?

Not simplifying, I would say make it more efficient (especially storage efficient) and more reasonable to store one sample table, then 3 sample tables in different configurations. I assume, now it will take 3x less space in pepdb.

nsheff:
> 

and will this also make it possible to use simplified projects, because they are unprocessed? like, I mean collapsed constant rows to the project level?

Khoroshevskyi:
> > 
> 
> and will this also make it possible to use simplified projects, because they are unprocessed? like, I mean collapsed constant rows to the project level?

Do you mean projects, such as geofetch is produsing (e.g. where config file contains append)?  - then Yes!

peppy/project.py
nsheff: What are you going to do about this?

nsheff:
> Do you mean projects, such as geofetch is producing (e.g. where config file contains append)? - then Yes!

Yes, that's what I mean. Great!


Can you rename this Pull Request to something more clear? Why does it say "402" ?

rafalstepien:
> > Do you mean projects, such as geofetch is producing (e.g. where config file contains append)? - then Yes!
> 
> Yes, that's what I mean. Great!
> 
> Can you rename this Pull Request to something more clear? Why does it say "402" ?

I think because it originated from issue number 402

nsheff:
> I think because it originated from issue number 402

Indeed. @Khoroshevskyi you have to use the # sign to link to the issue and indicate it as an issue number. Also, this probably belongs in the description of the PR

rafalstepien:
So after changing the name of PR are we good to go?

nsheff:
I made 1 request for a change, this still is not addressed (see above)

Khoroshevskyi:
> > Do you mean projects, such as geofetch is producing (e.g. where config file contains append)? - then Yes!
> 
> Yes, that's what I mean. Great!
> 
> Can you rename this Pull Request to something more clear? Why does it say "402" ?

As Rafal mentioned above, the name was originated from the issue number 402. All 4 changes are related to this problem. `eq` function is related to functionality `to_dict`; `from_dict` is dependent from `to_dict` ; And `from_pandas` has similar function as `from_dict` . That's why all this changes were made here. Is there something else to change @nsheff ?

nsheff (approved):

Khoroshevskyi:
I have chenged the names of the dict keys, could you check it please? @nsheff  --> https://github.com/pepkit/peppy/pull/405/commits/802ea1a112635f1f233a5faa12e0716bf13d8157

PR #409, opened 2022-08-30
Apply bugfix

### LINKED ISSUE #408: Peppy does not create correct list of `Sample()` objects stored in `self._samples`
Peppy does not create correct list of `Sample()` objects stored in `self._samples`. 

If there is a column in `sample_table.csv` or `subsample_table.csv` that is empty for any row, then `peppy` does not create `Sample()` object with this attribute. The problem is that this is not true, and `Sample()` instance should have this attribute with value `None`. To better explain:

Given following `sample_table.csv`:
```
| sample_name | fastq_1          | fastq_2          |
|-------------|------------------|------------------|
| name_1      | /path/to_fastq_1 | /path/to_fastq_2 |
| name_2      | /path/to_fastq_1 |                  |
| name_3      | /path/to_fastq_1 | /path/to_fastq_2 |
```

`peppy` will create the following structure:
```python
self._samples = [
    Sample(
        sample: name_1
        fastq_1:
            -  /path/to_fastq_1
        fastq_2:
            -  /path/to_fastq_2
    ),
    Sample(
        sample: name_2
        fastq_1:
            -  /path/to_fastq_1
    ),
    Sample(
        sample: name_3
        fastq_1:
            -  /path/to_fastq_1
        fastq_2:
            -  /path/to_fastq_2
    ),
]
```
And any attempt to access `self._samples[1].fastq_2` will raise an attribute error. In my opinion the correct structure should be:
```python
self._samples = [
    Sample(
        sample: name_1
        fastq_1:
            -  /path/to_fastq_1
        fastq_2:
            -  /path/to_fastq_2
    ),
    Sample(
        sample: name_2
        fastq_1:
            -  /path/to_fastq_1
        fastq_2:
            - None
    ),
    Sample(
        sample: name_3
        fastq_1:
            -  /path/to_fastq_1
        fastq_2:
            -  /path/to_fastq_2
    ),
]
```

rafalstepien:
All the reasons described in #408

closes #408


nleroy917 (approved):

nleroy917:
@rafalstepien this looks good to me. Wondering if we can bring in #410 with this? Should be only a line change.

vreuter (changes requested):
Hey thanks a lot for spotting an issue @rafalstepien , and even better for creating the changes to address it :)

I didn't yet look in detail, but one thing I noticed was that the test was in `smoketests`, but this is more like a logical test. I'm not sure the best placement since I've not been active in the project for a while, but maybe @nsheff or another recent contributor would have an idea for optimal placement of this new test among the others. Not a big deal, just will be easier to look up later and also better preserves the granularity in test execution, if we choose to run just a subset, when everything's well organized.

rafalstepien:
Sure, if there is any better place where you store unittests then let me know!

rafalstepien:
> @rafalstepien this looks good to me. Wondering if we can bring in #410 with this? Should be only a line change. This is useful to me in `pepingester` so I can set a default PEP version if one doesn't exist.
> 
> This actually will be helpful with `pephub` as well.

Sure, we can incorporate that.

rafalstepien:
@nsheff I would appreciate the review and decision what to do about test file (mentioned by @vreuter )

nsheff:
@rafalstepien I think this looks like some legacy organization, but it doesn't really make sense -- please just move any tests that you classify as "unit tests" outside of the "smoketests" folder? If there are tests in there that qualify as smoketests, then you can leave them there.

In the meantime, @stolarczyk can you clarify the distinction you were going for with putting these tests under "smoketests"? At first glance it looked like most of these are just unit tests.

nsheff:
@rafalstepien some updates for your example:

1. Examples should not be added directly to this repository, which is a read-only mirror of [example_peps](https://github.com/pepkit/example_peps). Please see: https://github.com/pepkit/peppy/blob/master/update_test_data.sh
2. If the above was not clear to you, can you please write instructions to this effect somewhere where you would have found them, for the next developer?
3. So you'll be adding a new example to that repo. Don't name that example "nextflow error" -- it isn't going to be an error in perpetuity. Come up with a name that makes sense for the example.

stolarczyk:
I decided to smoke test peppy to save time as we were rushing for a release at some point. 

To me, smoke testing means testing the entire `peppy.Project` initialization process and inspecting the result, which includes config parsing, preprocessing, applying the attribute modifiers etc. as opposed to testing individual functions  that perform these actions. Maybe that's an incorrect classification.

nsheff:
Thanks for the clarification @stolarczyk . I think that's fine.

So @rafalstepien -- please just classify your new tests as either smoketests (following michal's definition), in which case put them under `smoketests/`, or as unit tests, in which case you should remove them and put them in the `test` or under some subfolder other than `smoketests`.

rafalstepien:
So I moved most of the tests that in my opinion are unittests to apropriate file, removed redundancy from fixtures and renamed some example files. I will merge this after familiarizing with how to correctly add examples using the script you mentioned. Thank you all for help!

PR #411, opened 2022-08-31
use HTTPS links

fabianegli:
Make the internet a tiny bit safer.


nsheff:
Thanks for making these changes!

PR #413, opened 2022-09-01
Remove artifact converters 

### LINKED ISSUE #412: Remove artifact converters
There was a problem that reading sample_table to CSV sometimes returned in `NaN` values, that eg. @Khoroshevskyi  tried to solve by writing custom converters. I identified the root cause of this problem and introduced fix in https://github.com/pepkit/peppy/pull/409

Now we should go through `peppy` code and check if:
1) this actually solves the issue, if yes then:
2) remove unnecessary artifact functions, methods, converters that aimed to solved that problem, because it is not needed anymore

Khoroshevskyi:
Fixed #412


tests/smoketests/test_Project.py
```diff
@@ -22,21 +21,21 @@
...
+    # "duplicate",
+    # "imports",
+    # "subtable1",
+    # "subtable2",
+    # "subtable3",
+    # "subtable4",
     "subtable5",
-    "remove",
+    # "remove",
 ]
```
nsheff, line 40: are these comments supposed to be here?

PR #421, opened 2022-10-18
Add is_private attribute

rafalstepien:


nsheff:
you need to merge this into dev, so the master branch always reflects the latest release. this should have warranted a version bumpb

Issue #422, opened 2022-11-16
nan value appearing in return of to_dict when project was initiated from pandas 


Khoroshevskyi:
This issue already was solved #418
But it still appears, when project is initiated from pandas. e.g. using geofetch
```
a1 = Geofetcher().get_projects("GSE217424")["GSE217424_raw"]

aa1 = a1.to_dict(extended=True)["_sample_dict"]

```
`aa1` has in dictionary values `nan`:
![image](https://user-images.githubusercontent.com/41573628/202076307-c0a07f79-5b04-4f47-b924-e9ea93a60b9e.png)

Issue related to: https://github.com/pepkit/pepdbagent/issues/52

PR #423, opened 2022-11-16
Release 0.35.3

Khoroshevskyi:
This PR is a fix of #422 issue


nleroy917 (commented):
Have we confirmed this works? I know we were grappling with this for a while.

Khoroshevskyi:
Now, there is an error, when we want to create digest in pepdbagent, because nan values exists. I was checking this one line for 2 hours, and it works for me.

nleroy917:
Sounds good. LGTM.

PR #425, opened 2023-01-11
disabled progress bar

Khoroshevskyi:
Disabled progress bar that was printing while checking duplicated sample names.


nsheff:
remove dependency, imports, etc.

Khoroshevskyi:
Hi @stolarczyk , was there any specific reason you used track function while iterating list and other objects?
e.g.
https://github.com/pepkit/peppy/blob/29505052700cfa73068570ebc8899d6508c4366b/peppy/project.py#L473
https://github.com/pepkit/peppy/blob/29505052700cfa73068570ebc8899d6508c4366b/peppy/project.py#L490
https://github.com/pepkit/peppy/blob/29505052700cfa73068570ebc8899d6508c4366b/peppy/project.py#L506

stolarczyk:
That's solely to improve UX. It conditionally displays a progress bar for each sample modifier for large sample tables. 

If you're planning to remove it, it should have no downstream effects.

Khoroshevskyi:
> That's solely to improve UX. It conditionally displays a progress bar for each sample modifier for large sample tables.
> 
> If you're planning to remove it, it should have no downstream effects.

Ok, thanks! There was a bug, so this progress bar was printed always. I will fix it.

peppy/project.py
```diff
@@ -473,7 +474,7 @@ def _del_if_in(obj, attr):
             for s in track(
                 self.samples,
                 description="Removing sample attributes",
-                disable=not self.is_sample_table_large,
+                disable=not all([self.is_sample_table_large, self.progressbar]),
```
nsheff, line 477: instead of `all([self.is_sample_table_large, self.progressbar])` just  `self.is_sample_table_large and self.progressbar` seems more clear

nsheff (changes requested):
Just one minor change and then looks great!

peppy/project.py
Khoroshevskyi: Fixed

PR #426, opened 2023-01-18
switching print to sys.stderr in rich progressbar

Khoroshevskyi:
Added sys.stderr to rich progressbar


nsheff:
Don't merge into master, and connect to the other PR that this is extending.

I'm not that familiar with rich but this looks fine, otherwise.

nsheff (approved):

PR #428, opened 2023-02-03
Add link to the package on PyPI

vreuter:


nsheff:
Great! but is this link is pointing to the git ssh URI, not PyPI, or something?

vreuter:
Indeed, sorry, I must've typo'd on Ctrl+C when grabbing the URL from PyPI, and the repo clone was still in my buffer.

PR #438, opened 2023-03-23
from and to dict improvements

### LINKED ISSUE #437: Add functionality that supports non private keys in dictionary in `from_dict` fucntion
Few month ago I implemented `to_dict` and `from_dict` functionality to peppy. At that time I didn't know that we shouldn't return private variables (or dictionary with keys that start with underscore: eg. "_sample_dict"). This way of returning peppy as dictionary is incorrect and can lead to downstream problem. 

1. I suggest change this functionality, so to_dict will return all non-private keys.
2. from_dict will support dict with both private and non-private keys.

Khoroshevskyi:
In this PR I fixed #437 

Unfortunately, after this changes we have to solve one issue in pepdbagent.
Additionally I checked pephub, and it seams that everything works fine. 

@nleroy917, @nsheff please check if everything will work in pephub, before approval of this PR


tests/test_Project.py
```diff
@@ -561,7 +561,7 @@ def test_from_dict(self, example_pep_cfg_path):
         """
         p1 = Project(cfg=example_pep_cfg_path)
         p1_dict = p1.to_dict(extended=True)
-        del p1_dict["_config"]["sample_table"]
+        del p1_dict["config"]["sample_table"]
```
nleroy917, line 564: Should this be `CONFIG_DICT_KEY` ?

peppy/project.py
```diff
@@ -205,8 +206,10 @@ def from_dict(self, pep_dictionary: dict) -> "Project":
         """
         _LOGGER.info(f"Processing project from dictionary...")
 
+        pep_dictionary = {k.lstrip("_"): v for (k, v) in pep_dictionary.items()}
```
nleroy917, line 209: Is this needed now that we are removing all `_key`'s with `key`?

tests/test_Project.py
Khoroshevskyi: probably

peppy/project.py
Khoroshevskyi: Yes, because in pephub database all keys are with underscore (・へ・)

nleroy917: I know this is addressing issues with FastAPI. Would it make more sense to move this logic to the pephub server instead so that we aren't messing with core `peppy` stuff and potentially breaking things? Not sure.

I guess we could also just run a script on the database that does this conversion too so the PEPs there are compatible with the new `peppy`.

nleroy917 (commented):
The only thing I am weary of is this function that converts all underscore (`_`) prefixed keys to regular keys... Seems a bit hacky? I don't know. Maybe @nsheff can comment

nsheff (approved):

Khoroshevskyi:
After discussion we decided to not make changes in return dictionary (in to_dict func)

Issue #435, opened 2023-03-12
sample_modifiers applied to subsamples


johnyaku:
When a `subsample_table` is merged into the `sample_table` the new columns resulting from the merge are list columns.
This makes sense, as part of the motivation for subsample tables is to cater for situations where there a single sample has multiple values for a given property (column).

However, applying `sample_modifiers` such as `derive` breaks on these list columns.

More natural behaviour would be to iteratively apply the sample modifier rules to all elements in the list.
Alternatively, perhaps single item lists could be "unlisted" to become normal columns, and thus subject to `sample_modifiers`.

Not sure if this a feature request or a "how to" question...

johnyaku:
The motivation for wanting to be able to do this relates to issue #434.

Specifically, I want to be able to separate sample "properties" from "data paths".
I was hoping that putting one or the other in a subsample table would do the trick, but the single item lists are breaking things.

If this is actually a feature request (ie, there is no easy way to do this atm) then can I propose the idea of a "merge table", as similar to a subsample table in that it gets merged into the `sample_table` but where the expectation is that there will be no more than one row per sample.

johnyaku:
Digging deeper, it looks like that this is a feature request.

But a simple implementation (rather than introducing a `merge_table` item) would be to allow the specification of more than one `sample_table` as a list, as can now be done with `subsample_table`. There would need to be a common index, of course, but the tables could then be merged.

nsheff:
This makes a lot of sense to me, and I like the idea. I also agree that just making `sample_table` accept either a single table or a list of tables would give you the machinery that you need to separate your sample tables into two. It also wouldn't be too hard to implement.

But as I think about this more, I'm not certain this is better than the way I've typically approached this in the past: which is to use a combined append/derived attribute. With this approach, I simply don't put the "data paths" into the sample table at all, I only put them into the project config file. You can do 1) use `append` to create a new column; and then 2) use `derive` to turn it into a path.

This way, the paths are solely encoded in derived attribute modifiers, and the sample table has no paths. I discuss this a little bit here: 

https://pep.databio.org/en/latest/howto_eliminate_paths/

You can make it even more powerful in this example by *not* including the `file_path` in the sample table, but adding it in using `append` (or even using `imply` if it varies across the table).

johnyaku:
Thanks! Your suggestion works beautifully.

For anyone stumbling on this thread, my toy example looks like this:
```
sample_name,slide,area
CID_1234,V11Y24-111,A1
[...]

PR #440, opened 2023-03-27
Fix bug with custom sample id. See #435

nsheff:


vreuter:
Merging despite CI failure since it appears lint-related

PR #445, opened 2023-04-22
added handling of empty sample file

### LINKED ISSUE #444: If 'annotation_sheet.csv' is without samples, throws 'NoneType' error
I decided to remove the samples in the annotation_sheet.csv to see what would happen. Error was thrown - `TypeError: 'NoneType' object is not iterable.` It would be nice to have it respond with `No samples found in your file` or something.

ayobi:
fixes #444

### DIFF
--- a/peppy/project.py
+++ b/peppy/project.py
@@ -259,6 +259,9 @@ class Project(PathExAttMap):
         """
         self._samples: List[Sample] = self.load_samples()
         if modify:
+            if self.samples is None:
+                _LOGGER.info("No samples found in file.")
+                sys.exit(1)
             self.modify_samples()
         else:
             self._assert_samples_have_names()


peppy/project.py
```diff
@@ -259,6 +259,9 @@ def create_samples(self, modify: bool = False):
         """
         self._samples: List[Sample] = self.load_samples()
         if modify:
+            if self.samples is None:
+                _LOGGER.info("No samples found in file.")
+                sys.exit(1)
```
Khoroshevskyi, line 264: I would suggest to create new exception with message no samples found in the project for modification. And don't use sys.exit, because we are using peppy not only for files. We can have peppy as object, and sys.exit can cause issues in downstream projects.

Khoroshevskyi (changes requested):

peppy/project.py
ayobi: I see, ok, I will take it out. 
I made a mistake and put it in if modify condition, it should have been after samples were loaded but before modify. Just wanted to add a more helpful message in case their sample_annotation.csv file had no samples. I'll correct that. Thank you.

PR #448, opened 2023-05-25
Fixed #447 (Removal of attmap dependency)

### LINKED ISSUE #447: Remove attmap dependency
As in other projects, we should remove attmap here.

khoroshevskyi:
- Fixed #447 
- Removed OredredDict
- Deprecated Python<=3.7 
---

- To ensure back compatibility for `Sample` object, new simplified attmap superclass was created. [2ac4b02](https://github.com/pepkit/peppy/pull/448/commits/2ac4b02846bd86faf1f0e1c2a78fb9efadbda65c)
- Project class is simplified and it is not 100% back compatible (attriubtes and items are interchangeable: user can't set attriubte using `__setitem__`). But I think it is correct implementation.

 - [x] Solve merge conflicts
 - [x] Check from, to_dcit functionality, and if description and name is set correctly


docs/changelog.md
```diff
@@ -2,6 +2,12 @@
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format.
 
+## [0.35.6] -- 2023-05-25
+### Changed
+- Replaced attmap with MutableMapping. (which resulted in the removal of the attribute functionality previously available in attmap)
+- Replaced OrderedDict with dict.
+- Deprecated support for Python versions older than 2.7.
```
nsheff, line 9: is this 2.7 or 3.7?

peppy/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.35.5"
+__version__ = "0.35.6"
```
nsheff, line 1: Let's make this 0.40.0 since it's lost backwards compatibility.

peppy/sample.py
```diff
@@ -320,17 +301,37 @@ def project(self):
...
-    def __setattr__(self, key, value):
-        self._try_touch_samples()
-        super(Sample, self).__setattr__(key, value)
-
     def __delattr__(self, item):
         self._try_touch_samples()
         super(Sample, self).__delattr__(item)
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, item, value):
```
nsheff, line 308: why did you rename this from `key` to `item` ? I think `key` is more clear, I'd call 'item' the key:value pair.

peppy/_version.py
khoroshevskyi: 0.40.0 ? or 0.36.0?

peppy/sample.py
khoroshevskyi: I think, I was changing code few time and didn't mentioned this. You are right key is more clear

peppy/_version.py
nsheff: 0.40.0 since it's lost backwards compatibility.

PR #451, opened 2023-07-02
Add `numpy` explicitly to requirements.

mknorps:
This PR adds `numpy` library requirement explicitly to requirements-all.txt.

I am a [FawltyDeps](https://github.com/tweag/FawltyDeps) (dependency checker for Python projects) maintainer. I took the liberty to check `peppy` source code with:
```
fawltydeps --code peppy --deps requirements/requirements-all.txt 
```
and it looked great ❤️ I see you care a lot about keeping the code clean and reusable. I just found one small problem -  that `numpy` is not declared as a dependency and used explicitly in the code.

I would like to propose the following standard for 3rd party libraries declaration:
> If the 3rd party library is explicitly used in the code, it should be explicitly declared in the requirements file.

What do you think about it?

 `numpy` is usually installed along with `pandas`, but if for some reason you drop `pandas` or its underlying dependency changes, the code may not work.


nsheff (approved):

nsheff:
thanks for pointing this out, and fawltydeps sounds useful!

PR #484, opened 2024-05-20
Dev 459 sample table index

### LINKED ISSUE #459: the project missing `sample_name`, but with `sample_table_index` set correctly, cannot run in looper

donaldcampbelljr:
fixes #459

### DIFF
--- a/peppy/sample.py
+++ b/peppy/sample.py
@@ -302,6 +302,16 @@ class Sample(SimpleAttMap):
         """
         return self[PRJ_REF]
 
+    @property
+    def sample_name(self):
+        """
+        Get the sample's name
+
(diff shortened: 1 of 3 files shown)


khoroshevskyi (approved):
LGTM

peppy/sample.py
```diff
@@ -302,6 +302,16 @@ def project(self):
         """
         return self[PRJ_REF]
 
+    @property
+    def sample_name(self):
```
khoroshevskyi, line 306: Should it be named like `sample_name` or should it be different property name
Do you have a suggestion @nsheff ?

khoroshevskyi (approved):
Just one question, otherwise looks good

nsheff (approved):

PR #487, opened 2024-07-02
attempt to replace pandas with polars, this is untested

donaldcampbelljr:


khoroshevskyi:
This Pull request is old, and a lot of things changed. peppy will be overwritten in Rust and it will eliminate this problem

